PR #9, opened 2018-09-21
 bump  versions

doofin:


raquo:
This will be taken care of in #11. I generally update versions only when I'm about to release a new Airstream version for some other reason.

PR #11, opened 2018-10-14
New: Error handling

### LINKED ISSUE #7: Error Handling
As documented, currently an exception thrown anywhere in observables or observers is undefined behaviour. Obviously this is a non-starter for any production application, so the next version of Airstream will include error handling.

I haven't thought much about the specifics, so no special design decisions just yet, but in broad strokes I think it'll look something like this:

* the failure in an observable can be recovered from using methods like recover(err => value)
* the failure should be toxic (if unhandled, dependent observables should fail as well)
  * All of them, or just some? Does e.g. EventBus need an ErrorHandlingStrategy param?
* observers should be notified of errors with an optional onError
* it should be possible to print out all unhandled errors to dev console. Maybe a global stream exposing unhandled errors?

I will consider introducing observable completion at the same time, but no promises that I'll decide to implement it (see #1)

raquo:
- [x] Code
- [x] Tests
- [ ] Docs

Error handling in Airstream is unlike what you've seen in other streaming libraries. It is much more like what Scala.rx does, but adapted for our stricter execution model as well as Event Streams and Signals.

It behaves a lot more like plain Scala exception handling, except the errors are toxically propagating down the observable dependency graph instead of up the call stack.

Docs coming soon, with full explanation for how and why as usual.

Fixes #7


raquo:
If anyone has any feedback on this, now's the time. Docs for the changes are here: https://github.com/raquo/Airstream/blob/dabad308783f4d6e52fee946b48d8a8feee832d9/README.md#error-handling (added "Error Handling" section).

PR #12, opened 2018-10-30
adding filterNot to EventStream

yurique:
sometimes it's useful (and sometimes I'm lazy =) )

### DIFF
--- a/src/main/scala/com/raquo/airstream/eventstream/EventStream.scala
+++ b/src/main/scala/com/raquo/airstream/eventstream/EventStream.scala
@@ -36,7 +36,7 @@ trait EventStream[+A] extends LazyObservable[A] {
     new FilterEventStream(this, passes)
   }
 
-  def filterNot(passes: A => Boolean): EventStream[A] = filter(passes.andThen(!_))
+  def filterNot(predicate: A => Boolean): EventStream[A] = filter(!predicate(_))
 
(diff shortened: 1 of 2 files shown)


src/main/scala/com/raquo/airstream/eventstream/EventStream.scala
```diff
@@ -36,6 +36,8 @@ trait EventStream[+A] extends LazyObservable[A] {
     new FilterEventStream(this, passes)
   }
 
+  def filterNot(passes: A => Boolean): EventStream[A] = filter(passes.andThen(!_))
```
raquo: We'll instantiate one less Function if we just do `filter(!passes(_))`.

This might be a moot point after `fullOpt` but it can't hurt

raquo: Oh also, `passes` is no longer the right name, since its result indicates the opposite of passing.

Perhaps `skip`? Or maybe `predicate` if that doesn't feel right. I like having semantic names for such params, not just `f`...

yurique: renamed to `predicate`

yurique: also added a couple of tests (for map-flatten)

PR #25, opened 2019-07-24
New: `WriteBus.contracomposeWriter`

vic:
Allows people to provide an operator to transform the write-bus stream.

`contraCompose` is more general than `contramapWriter`, as the first takes
an operator on streams and the second a mapping function on its events.

The advantage of having `contraCompose` is that it provides a way for
people to combine streams.

Now `contramapWriter` is defined in terms of `contraCompose`.

### DIFF
--- a/src/main/scala/com/raquo/airstream/eventbus/WriteBus.scala
+++ b/src/main/scala/com/raquo/airstream/eventbus/WriteBus.scala
@@ -17,11 +17,15 @@ class WriteBus[A] extends Observer[A] {
     new EventBusSource(stream, sourceStream, owner)
   }
 
+  def contracomposeWriter[B](operator: EventStream[B] => EventStream[A])(implicit owner: Owner): WriteBus[B] = {
+    val mapBus = new WriteBus[B]
+    addSource(mapBus.stream.compose(operator))(owner)
+    mapBus
+  }
+
   /** Behaves similar to `contramap`, but gives you a WriteBus, not just an Observer */
   def contramapWriter[B](project: B => A)(implicit owner: Owner): WriteBus[B] = {
-    val mapBus = new WriteBus[B]
-    addSource(mapBus.stream.map(project))(owner)
-    mapBus
+    contracomposeWriter[B](_.map(project))(owner)
   }
 
   /** Behaves similar to `filter`, but gives you a WriteBus, not just an Observer */


vic:
Hey @raquo, thanks again for your work on Laminar and Airstream. Here's a PR for a function I'd find useful on WriteBus. My use case is, I have an `EventBus[Action]` I need to write to, but for creating the Action you'd need the input  (eg, the values produced from an input.onChange) and the current value of a signal. 

Something like this: 

```scala
case class Amount(amount: BigDecimal, currency: String)
case class State(playerName: String, balance: Amount)

sealed trait Action
case class SetAmount(amount: Amount) extends Action

val actionBus = ActionBus[Action]
val state = Signal[State]

// the resulting WriteBus can be given to anything expecting an Observer
// for example, something producing input.onChange events with the text value.
val inputObserver: Observer[String] = actionBus.contraCompose { (input$: EventStream[String]) =>
    // important thing here is you can combine input$ to produce an Action stream from it
    input$.map(BigDecimal).withCurrentValueOf(state).collect { 
       case (newAmount, state) if state.playerName == "Bob" =>
          SetAmount(state.balance.copy(amount = newAmount))
    }
}

```


Sidenote: Looks like the travis build is failing trying to install jdk 8, and expects at least version 9.

raquo:
Thanks @vic, this is great! Minimal and versatile.

I think we should rename it to either `composeWriter` or `contracomposeWriter`. Logically it should be the latter, however unlike "contramap", "contracompose" is not a term of art, in fact this issue one of only two mentions of it in both google and github (the universe is not as infinite as I thought huh).

Also if we went with `contracomposeWriter` we'd need to decide between that and `contraComposeWriter` (capital C), which I would feel better about, but would be inconsistent with `contramap` in `contramapWriter` being all lowercase.

For the record, `filterWriter` doesn't have `contra` in its name because its effect and type signature matches that of the expected `filter` operator, unlike contramap / contracompose which have the input / output types flipped.

So um, if anyone has thoughts on naming feel free to chime in. Otherwise I'll figure out which name I like more myself and will merge this on the weekend. I will look into travis at that time, thanks for the heads up.

Sidenote: I see you use trailing `$` in reactive variable names in your example code. Not to sound like a crazy old man, but if the ScalaJS compiler ever explodes in your face with weird internal errors, I'd blame that first. I wish I could be more specific but that was a while ago that I ran into this, and I don't remember the details anymore. I think it might have something to do with how the compiler encodes scala `object` names by adding trailing dollar signs to them, and possible name collisions arising from that (that are not apparent in the source code).

vic:
+1 for `contracomposeWriter` on my side.

Will rebase this branch and see if everything is green on CI. 

Regarding the trailing `$` haha, yeah. The ScalaJS compiler just went crazy with exactly that error today :D. Coming from cycle.js and xstream, using `$` was natural for me, I'll try to avoid it now.

raquo:
Haha, I figured. I ran into this issue due to cycle / xstream influence as well (I made and used scala.js facades for both a while ago). If you have time and if it's easy to reproduce, you can file an issue with scalajs since I'm pretty sure it's a legitimate bug. I was too lazy / short on time myself tbh.

PR #27, opened 2019-09-21
API: Allow users to override Owner method to create truly noop owners

raquo:
@cornerman Following up on our [gitter discussion](https://gitter.im/Laminar_/Airstream?at=5d85000cb9f8210ed5be0bc1) here's my proposed way to let you override the `own` method to make a custom `Owner` that is truly no-op.

In my current design, I need to keep the `own` method locked down to prevent end users from calling it, but I also need to allow the trait `Owned` (which is also in package `ownership`) to call it.

I don't think such an access modifier is possible with just one method, so instead I moved the implementation that you can override into one method, and made that method available to trait `Owned` via another method. I'm not super sure if this is good practice or a code smell. Please let me know if you have concerns.

### DIFF
--- a/src/main/scala/com/raquo/airstream/ownership/Owner.scala
+++ b/src/main/scala/com/raquo/airstream/ownership/Owner.scala
@@ -36,8 +36,10 @@ trait Owner {
     }
   }
 
-  private[ownership] def own(owned: Owned): Unit = {
+  protected[this] def _own(owned: Owned): Unit = {
     possessions.push(owned)
     onOwned(owned)
   }
+
+  @inline private[ownership] def own(owned: Owned): Unit = _own(owned)
 }


cornerman:
@raquo Thank you for looking into this. As I need a no-op owner and this change enables me to do it, I am happy :)

From the design perspective, I would love to not have any implementation in the Owner. Because as a library maintainer that wants to integrate with airstream, I need to know details about the default `Owner` implementation, in order to implement my desired behavior. In the case of no-op, I need to override `_own` (not necessarily `onOwned`) and maybe `onKilledExternally` (not necessarily `killPosessions`). And depending on the contract that we agree on for the owner, I need to check on updates whether I need to override more methods. Another use-case: I might want to keep subscriptions in a `Set` instead of an `Array`, which might be valid for certain patterns and numbers of subscribers.

For the access modifiers, I also do not think that there is such a modifier (doesn't java's `protected` behave like that? Subclasses and same package) :/

If end-user should be locked out, but depending on the use-case library-extensions or other parts of the library need to use it, I could think of a `private` token that proves that the call comes from a controlled environment:
```
sealed trait Token
object Token {
  private[ownership] implicit object token extends Token
}

// Then have methods depending on the token that you can only call internally
trait Owner {
  def own(owned: Owned)(implicit token: Token): Unit
}

// or you could even allow third-party libraries to create extensions which you run and pass in the token explicitly (this will get super sweet with scala 3 and implicit function types):
trait Extension[T](run: Token => T)
```

raquo:
Hey, sorry for the delay on this. I'm reworking quite a few things in Laminar / Airstream at the moment, including simplifying existing ownership concepts as well as expanding them with more features to solve Laminar's problems.

 I will have to re-evaluate this once the dust settles.

raquo:
Closing due to age, and the issue having been partially addressed. A lot of things changed since 2019 – if you're still interested in something like this, let's have a fresh conversation in a separate issue.

PR #30, opened 2020-02-02
New: Dynamic Ownership

raquo:
This will be v0.8 soon.

Dynamic Ownership is a pair of helpers to manage subscriptions that can be re-activated after being deactivated. There are no changes to how regular ownership works.

We actually had a similar dynamic system in pre-Airstream Laminar (v0.2?), but I took a wrong turn getting rid of that for the sake of supporting eager `State` observables. The latter failed to prove its worth and was removed from Airstream a while ago, so here we are, back to the right solution at last.

Using Dynamic Ownership in Laminar enabled me to fix the main memory management gotcha of Laminar, https://github.com/raquo/Laminar/issues/33 – I will publish that version soon.

Changelog lists other changes. Mostly just merging / simplifying the classes and types.


vic:
Awesome, @raquo. Thanks a lot for all your effort <3.

raquo:
@vic Thank you! Note it will take me a couple days to actually release this on Maven. I'll announce in gitter when it's actually out.

PR #40, opened 2020-10-06
Airstream v0.11.0 

### LINKED ISSUE #39: Glitch when updating a Var while reading its state
While Airstream solves vast majority of FRP glitches, a particular grief-causing pattern has emerged:

```scala
var n = 0
val clickBus = new EventBus[Unit]
val state = Var[List[Int]](Nil)
clickBus
  .events
  .map(_ => n = n + 2)
  .flatMap(_ => EventStream.seq(List(n -2, n - 1), emitOnce = true))
  .withCurrentValueOf(state.signal)
  .foreach { (next, currState) =>
    state.set(currState :+ next)
  }
}
clickBus.writer.onNext(())
println(state.now())
```

source: [gitter](https://gitter.im/Laminar_/Lobby?at=5f7239b75a56b467a5f7094c)

User expects state to contain `List(0, 1)` at the end, however in fact `state` will contain only `List(1)` because when the second event happens, `currState` will still be `Nil` because every `state.set` call will only update Var's state in a new transaction, which is created and thus scheduled AFTER both EventStream.seq transactions were created & scheduled, so it will also run AFTER they have completed.

To simplify, here is a similar pattern that results in the same issue for similar reasons:

```scala
var n = 0
val clickBus = new EventBus[Unit]
val state = Var[List[Int]](Nil)
clickBus
  .events
  .foreach { _ =>
    n = n + 2
    state.set(state.now() :+ (n - 2)
    state.set(state.now() :+ (n - 1)
  }
}
clickBus.writer.onNext(())
println(state.now())
```

source: also gitter, but older, can't find it now

In the example above, the call to `state.now()` will return `Nil` both times, because the transaction kicked off by `state.set(state.now() :+ (n - 2)` hasn't processed yet, because the current transaction hasn't finished yet.

One solution could be to replace `state.set(state.now() :+ (n - 1)` with `new Transaction(_ => state.set(state.now() :+ (n - 1))`, which would delay evaluation of `state.now()` until a new transaction which is scheduled after the transaction created by `state.set(state.now() :+ (n - 2)`. However, this is ugly both visually and conceptually. This call will actually create two transactions. Maybe we can add a `nowSync(currValue => ())` to Var which would trigger the callback only after the current transaction or even ALL transactions have been cleared?

Alternatively, we already have `update(mod: A => A)` method on Var, however it evaluates the next state immediately when called. Perhaps it should instead delay the evaluation until the transaction that it creates actually starts executing.

Currently we have:

```
def update(mod: A => A): Unit = {
    val unsafeValue = now()
    val nextValue = Try(mod(unsafeValue))
    new Transaction(setCurrentValue(nextValue, _))
  }
```

It would be pretty simple to move `unsafeValue` and `nextValue` evaluation inside the Transaction callback.

Yet another alternative is for the user to introduce an asynchronous delay somewhere between two events that they fire. So add `delay(0)` to `EventStream.seq(List(n -2, n - 1), emitOnce = true)` in the first example, or wrap `state.set(state.now() :+ (n - 1)` in `js.timers.setTimeout` in the second one.

Also I should document this issue and workarounds more clearly in Laminar & Airstream docs.

If anyone has other ideas or concerns about proposed solutions, please speak up. This issue will likely require breaking changes to the API where it will still compile but the timing logic would be different, requiring manual review of your usages of Var.update and such.

raquo:
Allegedly fixes #39.

First version available as `0.11.0-M1`. Documentation not updated yet.

I will add more unrelated changes to this branch later.

---

What I did to fix the glitch:

var.update, Var.update and the corresponding try* methods will evaluate the provided `mod` callbacks only when their transaction is executed, not immediately as before. That was obvious, but not nearly enough on its own. The complex part is this:

I changed transaction scheduling logic. Transactions are scheduled when you create a new transaction – e.g. call var.set or var.update or eventbus.writer.onNext – while a transaction is already running, so e.g. inside an `observable.foreach` block. This part didn't change, that's just the definition.

Previously, those scheduled transactions were then executed in the order in which they were scheduled. In this PR I changed this. Instead, if you for example have scheduled transactions B and C while transaction A was running, transaction B will execute after A, but here's what happens after that: if you have scheduled any transactions while B was running, those transactions will execute before transaction C is executed, even though they were scheduled later than C. And this is recursive. This way, transaction B can properly and completely "finish" (even if it requires more transactions), before transaction C, which was intended to go after it, starts executing.

If what I'm saying is not clear, please see the last test in GlitchSpec.scala

I think this new logic makes more sense, and it's no coincidence that it also happens to fix the glitch. The previous approach resulted in surprising and somewhat arbitrary timing behaviour as demonstrated by #39.

---

I published this as a milestone release so that you can test it on your codebase. Especially if you ran into #39, but even if not, this change in behaviour might still affect you if you somehow relied on previous transaction timing. **Please throw out the `delay(0)` / `setTimeout` workarounds when testing, this should work without them.** This branch will become Airstream v0.11.0, so please check this new behaviour and let me know if you needed to fix anything or if it works with no adjustments. All the Laminar tests pass, and my own private codebase works just fine (after a lot of trial and error getting the queue algorithm right) so I'm optimistic.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,12 @@
-# Changelog
+# Changelog (pre-v0.11.0)
 
 Breaking changes in **bold**.
 
+---
+
+#### For versions v0.11.0 and up, see [laminar.dev/blog](https://laminar.dev/blog)
+
+---
 
 #### v0.10.2 – Sep 2020
 
--- a/README.md
+++ b/README.md
@@ -18,10 +18,10 @@ Airstream is a small state propagation and streaming library. Primary difference
 
 Airstream has a very generic design, but is primarily intended to serve as a reactive layer for unidirectional dataflow architecture in UI components. As such, it is not burdened by features that cause more problems than they solve in frontend development, such as backpressure and typed effects.
 
-I created Airstream because I found existing solutions were not suitable for building reactive UI components. My original need for Airstream was to replace the previous reactive layer of [Laminar](https://github.com/raquo/Laminar), but I'll be happy to see it used by other reactive UI libraries as well. Another piece of Laminar you can reuse is [Scala DOM Types](https://github.com/raquo/scala-dom-types).
+I created Airstream because I found existing solutions were not suitable for building reactive UI components. My original need for Airstream was to replace the previous reactive layer of [Laminar](https://laminar.dev), but I'll be happy to see it used by other reactive UI libraries as well. Another piece of Laminar you can reuse is [Scala DOM Types](https://github.com/raquo/scala-dom-types).
 
 ```
-"com.raquo" %%% "airstream" % "0.10.2"  // Scala.js 1.x only
+"com.raquo" %%% "airstream" % "0.11.0"  // Scala.js 1.x only
 ```
 
 
@@ -41,6 +41,8 @@ I created Airstream because I found existing solutions were not suitable for bui
   * [Observer](#observer)
   * [Ownership](#ownership)
     * [Ownership & Memory Management](#ownership--memory-management)
+    * [OneTimeOwner](#onetimeowner)
+    * [ManualOwner](#manualowner)
     * [Dynamic Ownership](#dynamic-ownership)
   * [Sources of Events](#sources-of-events)
     * [Creating Observables from Futures and Promises](#creating-observables-from-futures)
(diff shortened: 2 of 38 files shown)


vic (dismissed):
Tested this branch on my codebase, the glitch disappears, no need for the `.delay(0)` workaround. 

Thank you very much, @raquo

raquo:
Discussion of transaction infinite loops: [gitter](https://gitter.im/Laminar_/Lobby?at=5f7e18bf66ee4e44a88b3a80)

PR #48, opened 2020-12-25
Next: v0.12

raquo:


raquo:
FYI I moved around most of the classes in Airstream. The previous package structure was really annoying to navigate, at least for me when working on stuff.

This change might be annoying to users, but hopefully not terribly so, as I kept the actual class names the same, and most classes for most users will be imported via aliases on Laminar's `L` object.

However, this change also means that you certainly can't use this unpublished version of Airstream with the current version of Laminar using publishLocal anymore. Hang tight, I'll update Laminar for this soon.

ngbinh:
@raquo I am wondering what left for `0.12`?

raquo:
@ngbinh In terms of features https://github.com/raquo/Airstream/issues/61 and https://github.com/raquo/Airstream/issues/56 (the latter in Laminar). Plus all the regular chores of writing the docs (only CustomSource is left), the change log, updating Waypoint, etc.

I hope to publish 0.12.0-m1 of Waypoint / Laminar / Airstream for you guys to check out some time this week, followed by 0.12.0 after a few days if nothing explodes.

ngbinh:
Cool, thanks for the update

src/main/scala/com/raquo/airstream/util/JsPriorityQueue.scala
```diff
@@ -2,23 +2,22 @@ package com.raquo.airstream.util
 
 import scala.scalajs.js
 
-// @TODO[Test] add tests for this class
```
raquo, line 5: Well, I'm doing it now!

src/main/scala/com/raquo/airstream/combine/generated/CombinableSignal.scala
```diff
@@ -0,0 +1,253 @@
...
+import com.raquo.airstream.core.Signal
+
+// #Warning do not edit this file directly, it is generated by GenerateCombinableSignal.scala
+
+// These combine / combineWith / withCurrentValueOf / sample methods are implicitly available on all signals
+// For combine / combineWith methods on the Signal companion object, see StaticSignalCombineOps.scala
+
+class CombinableSignal[A](val signal: Signal[A]) extends AnyVal {
+
+  def combineWith[T1](
```
kitlangton, line 13: Just a boring lil' naming suggestions:

Have you considered `zip` for this one? And `zipWith` for `combineWithFn`? 
I like the terseness/analogousness to List's zip. But I can live either way! :D

This update looks awesome btw!

src/main/scala/com/raquo/airstream/combine/generated/StaticEventStreamCombineOps.scala
```diff
@@ -0,0 +1,148 @@
...
+import com.raquo.airstream.core.EventStream
+
+// #Warning do not edit this file directly, it is generated by GenerateStaticEventStreamCombineOps.scala
+
+// These combine and combineWith methods are available on the EventStream companion object
+// For instance methods of the same name, see CombinableEventStream.scala
+
+object StaticEventStreamCombineOps {
+
+  def combine[T1, T2](
```
kitlangton, line 12: Then you could use `zip` for here as well :D I see why combineWith may be a lil' strange to have as a companion function.

kitlangton (commented):
Throwing nitpicks from the shadows!
Great job @raquo!

src/main/scala/com/raquo/airstream/combine/generated/CombinableSignal.scala
yurique: I'd love to have `zip`, too. 

But it appears that `zip` has different semantics from what we have here (see Nikita's comment [here](https://github.com/raquo/Airstream/pull/51#issuecomment-751306016)).

(I, personally, could live with this inconsistency vs RxJS (at least for Signals) :) )

raquo: As explained in the linked comment, zip has legitimately different semantics from combine, it's not just a feature of ReactiveX, Monix and all other streaming libraries I know behave the same way. If we had an operator called zip that did a different thing than what's standard, many people would consider it a bug.

On lists, "zip" seems like the best match to what we're doing only because combineLatest can't exist on a list, since there is no concept of time in a list.

Issue #47, opened 2020-12-25
Improve/replace combineWith


ajaychandran:
Consider `EventStream.combineWith`:
```scala
  def combineWith[AA >: A, B](otherEventStream: EventStream[B]): EventStream[(AA, B)] = {
    new CombineEventStream2[AA, B, (AA, B)](
      parent1 = this,
      parent2 = otherEventStream,
      combinator = CombineObservable.guardedCombinator((_, _))
    )
  }

// usage
case class Point(x: Int, y: Int)
val $x: EventStream[Int] = ???
val $y: EventStream[Int] = ???
val $p: EventStream[Point] = $x.combineWith($y).map2(Point)
```
This combinator can be replaced with the more generalized form:
```scala
// simply exposing the additional parameter avoids unnecessary tuple allocation
// also, type parameter AA is redundant
  def zipWith[B, C](that: EventStream[B])(f: (A, B) => C): EventStream[C] =
    new CombineEventStream2(
      parent1 = this,
      parent2 = that,
      combinator = CombineObservable.guardedCombinator(f)
    )

// replacement for combineWith
  def zip[B](that: EventStream[B]): EventStream[(A, B)] =
    zipWith(that)(_ -> _)

// usage
case class Point(x: Int, y: Int)
val $x: EventStream[Int] = ???
val $y: EventStream[Int] = ???
val $p: EventStream[Point] = $x.zipWith(y)(Point)

```
What do you think?

raquo:
Nice 👍 . Also reduces the number of observables by half, since `map2(Point)` is not needed.

I don't want to use `zip` name because Airstream's `combineWith` doesn't work like `zip` does in ReactiveX, it doesn't wait for all of its inputs to produce events before emitting, it's much closer to `combineLatest`.

I think renaming the original method `combine` and naming the new method `combineWith` would be fine, except for the migration pain. In my own app migration would be trivial. Does anyone have concerns? I'm inclined to implement this.

Alternatively, I thought about overloading `combineWith` with an extra combinator param in the first argument list, but while it works for case classes, it requires types ascribtions if combinator is an anonymous lambda. And omitting ascribtions would be convenient for example if you wanted to do two subsequent combine-s and get an `EventStream[Tuple3[A, B, C]]`: then you'd need to provide an anonymous lambda to massage `(A, B), C` into `(A, B, C)`.

yurique:
[...]

PR #51, opened 2020-12-25
Expand combine, combineWith, etc. to support N arguments, etc.

yurique:
@raquo does `Signal.seq` implementation look right or can it be implemented more efficiently?

Also, included is a source code generator that generates the `zip` functions for the `Signal` companion object, to be used like this:

```
Signal.zip(signal1, signal2, signal3) // : Signal[(T1, T2, T3)]
```

The generated classes look like this:

```scala
package com.raquo.airstream.signal

import com.raquo.airstream.core.AirstreamError.CombinedError
import com.raquo.airstream.features.{ CombineObservable, InternalParentObserver }
import scala.util.{ Failure, Success, Try }

class ZipSignal3[T1, T2, T3](
  protected[this] val parent1: Signal[T1],
  protected[this] val parent2: Signal[T2],
  protected[this] val parent3: Signal[T3]
) extends Signal[(T1, T2, T3)] with CombineObservable[(T1, T2, T3)] {

  override protected[airstream] val topoRank: Int = (
    parent1.topoRank max parent2.topoRank max parent3.topoRank
  ) + 1

  parentObservers.push(
    InternalParentObserver.fromTry[T1](parent1, (nextValue, transaction) => {
      internalObserver.onTry(guarded(nextValue, parent2.tryNow(), parent3.tryNow()), transaction)
    }),
    InternalParentObserver.fromTry[T2](parent2, (nextValue, transaction) => {
      internalObserver.onTry(guarded(parent1.tryNow(), nextValue, parent3.tryNow()), transaction)
    }),
    InternalParentObserver.fromTry[T3](parent3, (nextValue, transaction) => {
      internalObserver.onTry(guarded(parent1.tryNow(), parent2.tryNow(), nextValue), transaction)
    })
  )

  private def guarded(
    t1: Try[T1],
    t2: Try[T2],
    t3: Try[T3]
  ) = {
    if (Seq(t1, t2, t3).forall(_.isSuccess)) {
      Success((t1.get, t2.get, t3.get))
    } else {
      Failure(CombinedError(Seq(t1, t2, t3).map(_.failed.toOption)))
    }
  }

  override protected[this] def initialValue: Try[(T1, T2, T3)] =
    guarded(parent1.tryNow(), parent2.tryNow(), parent3.tryNow())

}
```

`SignalZips` trait looks like this:

```scala

trait SignalZips {

  @inline
  def zip[T1, T2](
    s1: Signal[T1], 
    s2: Signal[T2]
  ): Signal[(T1, T2)] = new ZipSignal2[T1, T2](s1, s2)

  @inline
  def zip[T1, T2, T3](
    s1: Signal[T1], 
    s2: Signal[T2], 
    s3: Signal[T3]
  ): Signal[(T1, T2, T3)] = new ZipSignal3[T1, T2, T3](s1, s2, s3)

// ...
```

And a spec is generated as well (with test cases for each of the functions).

This can be controlled in the build.sbt:

```
Compile / sourceGenerators += Def.task {
  Seq.concat(
    (2 to 22).flatMap(n => // generate ZipSignalN for n from 2 to 22
      new ZipSignalGenerator((Compile / sourceManaged).value, n).generate()
    ), 
    new SignalZipsGenerator((Compile / sourceManaged).value, from = 2, to = 22).generate() // generate Signal.zip() for 2 to n signals
  )
}.taskValue

Test / sourceGenerators += Def.task {
  Seq.concat(
    (2 to 22).flatMap(n => // generate tests for Zip(s1, s2) ... Zip(s1, ..., s22)
      new ZipSignalTestGenerator((Test / sourceManaged).value, from = 2, to = 22).generate()
    )
  )
}.taskValue
```

The generated code can be found here:

* code - `Airstream/target/scala-2.13/src_managed/main/scala/com/raquo/airstream/signal/`
* tests — `Airstream/target/scala-2.13/src_managed/test/scala/com/raquo/airstream/signal/`

```

### CODE REVIEW

File: src/main/scala/com/raquo/airstream/signal/SeqSignal.scala [hunk 1]
```diff
@@ -0,0 +1,51 @@
...
+import com.raquo.airstream.features.{ CombineObservable, InternalParentObserver }
+
+import scala.util.{ Failure, Success, Try }
+
+class SeqSignal[A](
+  protected[this] val parents: Seq[Signal[A]]
+) extends Signal[Seq[A]]
+    with CombineObservable[Seq[A]] {
+
+  override protected[airstream] val topoRank: Int = parents.map(_.topoRank).maxOption.fold(0)(_ + 1)
```
raquo: Minimal topoRank should be 1, not 0

raquo:
I like the idea of generating N classes. I guess with enough .asInstanceOf-s your SeqSignal could be used for all that instead, but properly typed individual files are better.

We shouldn't use the "zip" name for this because [zip](https://rxmarbles.com/#zip) has different semantics than our `combineWith` (and your `zip`), which is much closer to [combineLatest](https://rxmarbles.com/#combineLatest) behaviour. Same for the other issue – we're not talking about introducing a new zip operator, it's all about the same old combines, just with slightly different method signatures for convenience.

Regarding SeqSignal, I think it's nice to have, and if we're adding that we'd need to add one for streams too, but I want to sort out all the other combine stuff first.

yurique:
I've updated the PR with the following...

### renamed `zip` to `combine`

```scala
private[signal] trait SignalCombines {

  @inline
  def combine[T1, T2, T3](
    s1: Signal[T1], 
    s2: Signal[T2], 
    s3: Signal[T3]
  ): Signal[(T1, T2, T3)] = new CombineSignal3[T1, T2, T3](s1, s2, s3)

// ...
}
```

```scala
private[eventstream] trait EventStreamCombines {

  @inline
  def combine[T1, T2, T3](
    s1: EventStream[T1], 
    s2: EventStream[T2], 
    s3: EventStream[T3]
  ): EventStream[(T1, T2, T3)] = new CombineEventStream3[T1, T2, T3](s1, s2, s3)

// ...
}
```

```scala
object Signal extends SignalCombines {
// ...

// for 2-tuple 
  @inline
  def combine[T1, T2](s1: Signal[T1], s2: Signal[T2]): Signal[(T1, T2)] = s1.combine(s2)

// ...
}
```

```scala
object EventStream extends  EventStreamCombines {
// ...

// for 2-tuple 
  @inline
  def combine[T1, T2](s1: Signal[T1], s2: Signal[T2]): Signal[(T1, T2)] = s1.combine(s2)

// ...
}
```


###  `CombineNSignal[A, Out]` and `CombineNEventStream[A, Out]`

These behave like the previous `SeqSignal`, but with an additional 

```scala
  protected def toOut(seq: Seq[A]): Out
```

`SeqSignal` and `SeqEventStream` extend these, with `toOut` = `identity`:


```scala
class SeqSignal[A](
  parents: Seq[Signal[A]]
) extends CombineNSignal[A, Seq[A]](parents) {

  override protected def toOut(seq: Seq[A]): Seq[A] = seq

}
```


```scala
// probably needs a better name: tried SeqEventStream, but someone beat me to it
class SeqJoinEventStream[A](
  parents: Seq[EventStream[A]]
) extends CombineNEventStream[A, Seq[A]](parents) {

  override protected def toOut(seq: Seq[A]): Seq[A] = seq

}

```

```scala
object EventStream { 
// ...
  def seq[A](streams: Seq[EventStream[A]]): EventStream[Seq[A]] = new SeqJoinEventStream[A](streams)
// ...
}
```

```scala
object Signal { 
// ...
  @inline def seq[A](signals: Seq[Signal[A]]): Signal[Seq[A]] = SeqSignal[A](signals)
// ...
}
```

The generated `CombineSignal{n}` and `CombineEventStream{n}` look like this:

```scala
class CombineSignal4[T1, T2, T3, T4](
  protected[this] val parent1: Signal[T1],
  protected[this] val parent2: Signal[T2],
  protected[this] val parent3: Signal[T3],
  protected[this] val parent4: Signal[T4]
) extends CombineNSignal[Any, (T1, T2, T3, T4)](
  Seq(parent1, parent2, parent3, parent4)
){

  def toOut(seq: Seq[Any]): (T1, T2, T3, T4) = 
    (seq(0), seq(1), seq(2), seq(3)).asInstanceOf[(T1, T2, T3, T4)]

}
```

```scala
class CombineEventStream4[T1, T2, T3, T4](
  protected[this] val parent1: EventStream[T1],
  protected[this] val parent2: EventStream[T2],
  protected[this] val parent3: EventStream[T3],
  protected[this] val parent4: EventStream[T4]
) extends CombineNEventStream[Any, (T1, T2, T3, T4)](
  Seq(parent1, parent2, parent3, parent4)
){

  def toOut(seq: Seq[Any]): (T1, T2, T3, T4) = 
    (seq(0), seq(1), seq(2), seq(3)).asInstanceOf[(T1, T2, T3, T4)]

}
```

### tuple composition

I copied the code generators from `tuplez` and implemented the changes suggested in https://github.com/raquo/Airstream/issues/47 :

```scala
class Signal {
// ...
  def combineWith[B, C](otherSignal: Signal[B])(project: (A, B) => C): Signal[C] = {
    new CombineSignal2[A, B, C](
      parent1 = this,
      parent2 = otherSignal,
      combinator = CombineObservable.guardedCombinator(project)
    )
  }

  @inline def combine[B](otherSignal: Signal[B])(implicit composition: Composition[A, B]): Signal[composition.Composed] = {
    combineWith(otherSignal)(composition.compose)
  }
// ...
}
```

```scala
class EventStream {
// ...

  def combineWith[B, C](otherEventStream: EventStream[B])(project: (A, B) => C): EventStream[C] = {
    new CombineEventStream2[A, B, C](
      parent1 = this,
      parent2 = otherEventStream,
      combinator = CombineObservable.guardedCombinator(project)
    )
  }

  @inline def combine[B](otherEventStream: EventStream[B])(implicit composition: Composition[A, B]): EventStream[composition.Composed] =
    combineWith(otherEventStream)(composition.compose)

  // "naming things" has defeated me this time :)
  def withCurrentValueOfWith[B, C](signal: Signal[B])(project: (A, B) => C): EventStream[C] = {
    new SampleCombineEventStream2[A, B, C](
      samplingStream = this,
      sampledSignal = signal,
      combinator = CombineObservable.guardedCombinator(project)
    )
  }

  @inline def withCurrentValueOf[B](signal: Signal[B])(implicit composition: Composition[A, B]): EventStream[composition.Composed] = {
    withCurrentValueOfWith(signal)(composition.compose)
  }
// ...
}
```

### note on composition

I respect the goal of keeping Airstream dependency-less :) and I'd love to see these changes get in even if we keep the tuple-related stuff in the Airstream codebase. 

But my 2 cents are:

* `tuplez` is a small lib with no dependencies itself, and will barely need to ever change (except maybe porting to scala-3 and releasing for scala.js v2 some day)
* I use it heavily in frontroute — and there are things in tuplez that are not needed for Airstream (I haven't copied the generators for those), which means I'll probably need to extract those into frontroute and drop the tuplez dependency as well (because half of what I need will be in Airstream)

This means there will be three projects that generate subsets of the same thing:
AIrstream (tuple composition), frontroute (apply converters) and tuplez (both composition and apply converters).

So I'd be more happy if we used tuplez everywhere.

BUT: I will still be EXTREMELY happy with how things are in this PR :).

### WIP

Please don't hesitate to point out things that need to change (including code-style, naming, formatting, etc) — I'm here to do that ASAP.

raquo:
I admit I am also annoyed about including all this tuple stuff that should have been in Scala stdlib into Airstream. But at the same time, tuplez has thousands of lines of generated code and more than 300 implicits. I totally see why these are needed to achieve tuplez' goal, and I definitely see that conceptually it's all pretty simple, but I just don't want to include that much stuff in Airstream. I don't really care that users won't be able to combine Tuple10 with Tuple12, I'd rather add support for combining I dunno at most 8-10 values, and if someone has crazy use cases they can create or import additional stuff from somewhere. Just wanted to elaborate on my thoughts on this.

So anyway, I checked out tuplez to look at it in more details, and I have a few questions about it. To start with:

- Are Tuple and NonTuple types and evidences really necessary? At least for Airstream use cases, I would think that you could manage priority by depioritizing Composition-s that have scalars on one side or the other (using trait hierarchy), and everything would still work similar to how I described in [this comment](https://github.com/raquo/Airstream/issues/47#issuecomment-751310697).

- Why are Composition_Pri10 and Composition_Pri20 separate traits? Will something break if they are combined?

yurique:
> Are Tuple and NonTuple types and evidences really necessary?

`Tuple` is just a provided utility, not used by the `tuplez` itself and doesn't have to be there (I wonder if I overestimated its usefulness). `Tuple` stuff can be definitely removed.

`NonTuple` is used to disambiguate the implicits.

> Why are Composition_Pri10 and Composition_Pri20 separate traits? Will something break if they are combined?

You're right, it's not necessary (it's a leftover from a previous version).

After looking more carefully into it, it seems that `NonTuple` is not necessary, either.

So now there will be 6 modules published:

* `tuplez-full`  — tupleN + scalar, scalar+tupleN, tupleN+tupleM, for all sizes (2 to 22) ([gist](https://gist.github.com/yurique/2e465979e89bb86214a4ccb6a0adc66c))
* `tuplez-full-light`  — tupleN + scalar, scalar+tupleN, tupleN+tupleM, for all sizes (2 to 10)([gist](https://gist.github.com/yurique/80f8fb8701e49cfd749348015f1f546a))
* `tuplez-basic`  — tupleN + scalar, for all sizes (2 to 22) ([gist](https://gist.github.com/yurique/6f53c0da31b30c6f637b6292aedfd144))
* `tuplez-basic-light`  — tupleN + scalar, for all sizes (2 to 10) ([gist](https://gist.github.com/yurique/482ca61973266dead614367b5ee833f7))
* `tuplez-apply` — something I use in frontroute, not needed for Airstream
* `tuplez-tuple` — `Tuple` evidences (I also use it in frontroute, also not needed for Airstream)

yurique:
So we can either:
* add a dependency 
* I can copy the generator into this PR

I'm good with both of these options :)

For me — basic-light (or maybe basic so users don't get any surprises) seems to be enough. It looks like `Signal.combine` / `Stream.combine` will be a better fit in most of the cases.

yurique:
Another idea crossed my mind — we can have Airstream depend on a very minimal version of tuplez, which would only provide the `Composition` trait itself and this:

```scala
abstract class Composition[-A, -B] {
  type Composed
  val compose: (A, B) => Composed
}

object Composition extends Composition_Pri10 {
  implicit def ***[A, B] = Composition[A, B, (A, B)]((_, _))
  type Aux[A, B, O] = Composition[A, B] { type Composed = O }

  def apply[A, B, O](c: (A, B) => O): Aux[A, B, O] =
    new Composition[A, B] {
    override type Composed = O
    val compose: (A, B) => O = c
  }

}
```

and users will be able to then swap it to the version they need (basic, full, -light etc) by removing the transitive dependency and adding another one like this:

```scala
libraryDependencies += ("com.raquo" %%% "laminar" % "0.11.0").excludeAll(ExclusionRule(organization = "app.tulz"))
libraryDependencies += "app.tulz" %%% "tuplez-full" % "0.3.0"
```

raquo:
@yurique Thanks for clearing up a bunch of stuff! This is looking great, especially `tuplez-full-light`.

So if Airstream will depend on `tuplez-full-light`, but the user will want `tuplez-full` because they have 22-element tuples for unspeakable reasons, are you saying the user can do that `excludeAll` thing, and they'll be able to use everything from `tuplez-full`? And Airstream could internally use some stuff from `tuplez-full-light`, and that would work too? Sounds a bit crazy, but honestly that would be a good compromise. How does that work though, do you need to ensure that the `tuplez-full` artifact is source and binary compatible with `tuplez-full-light` or something?

Also, what are your thoughts on using path dependent types (`composition.Composed`) vs adding a type param for the output? I understand that an extra type param will make type signatures 50% longer but honestly they're kind of a writeoff already. I'm just a bit worried about potential issues with path dependent types. Don't even know what exactly, stuff like compilation performance or type inference...

raquo:
PS sorry I'm a bit slow in replying, I'm in too deep trying to make custom stream integrations easier (prompted by the ajax thing)

yurique:
@raquo 
Without path dependent types this won't work (type inference won't, to be precise).

If we had this:

```scala
abstract class Composition[-A, -B, Composed] {
  val compose: (A, B) => Composed
}

// .. same prioritized instances
// and 

object Composition extends Composition_Pri10 {
// ...
  def apply[A, B, O](c: (A, B) => O): Composition[A, B, O] =
    new Composition[A, B, O] {
      val compose: (A, B) => O = c
    }
}
```

and a function like this:

```scala

object Test {
  def compose[L, R, O](l: L, r: R)(implicit c: Composition[L, R, O]): O =
     c.compose(l, r)
```

Then this woudn't compile:
```scala
  // scalac knows the types for L and R, but O is free, thus it fails to resolve things
  compose((1, 2), 3) // could not find implicit value for parameter c: Composition[(Int, Int),Int,O]
```

(although this would:
```scala
  compose((1, 2), 3)(Composition.`T2+scalar`)
```
)

yurique:
As for the swapping of the dependencies: I believe it works as long as the libs are binary- (or "sjs-"?) compatible. And that seems to be the case in my little experiment.

I publish-local'ed the version of Airstream that depends on tuplez-full-light. 

The code:
```scala
  {
    val s1: Signal[(Int, Int, Int, Int, Int)] = ???
    val s2: Signal[(String, String, String, String, String)] = ???

    val s3: Signal[(Int, Int, Int, Int, Int, String, String, String, String, String)] =
      s1.combine(s2)
  }
  {
    val s1: Signal[(Int, Int, Int, Int, Int, Int, Int, Int, Int, Int)] = ???
    val s2: Signal[(String, String, String, String, String, String, String, String, String)] = ???

    val s3: Signal[(Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, String, String, String, String, String, String, String, String, String)] =
      s1.combine(s2)
  }  
```

when I have `libraryDependencies += "com.raquo"   %%% "airstream" % "0.11.2-SNAPSHOT"` it fails to compile the second block (first one compiles):

```
[error]  found   : com.raquo.airstream.signal.Signal[((Int, Int, Int, Int, Int, Int, Int, Int, Int, Int), String, String, String, String, String, String, String, String, String)]
[error]  required: com.raquo.airstream.signal.Signal[(Int, Int, Int, Int, Int, Int, Int, Int, Int, Int, String, String, String, String, String, String, String, String, String)]
[error]       s1.combine(s2)
```
(it used the low priority composition instance)

But when I did this:
```scala
  libraryDependencies += ("com.raquo"  %%% "airstream" % "0.11.2-SNAPSHOT").excludeAll(ExclusionRule(organization = "app.tulz"))
  libraryDependencies += "app.tulz" %%% "tuplez-full"  % "0.3.0"
```

— it successfully compiled both blocks. And bsp reports only `tuplez-full` as a dependency (`tuplez-full-light` is removed as intended).

raquo:
@yurique Cool, cool. I finished my ajax / custom sources stuff, so I'll get back to tuplez stuff tomorrow.

raquo:
@yurique To your prior question, now that I looked into it: yes, I think I can make CombineNEventStream / *Signal performance quite a bit better, and eliminate the purity requirement for the combinator function at the same time. Working on it now.

raquo:
@yurique FYI I'm gonna be rebasing / force pushing this a lot in the next few hours

raquo:
Alright, I've done a bunch of stuff. EventStream companion object now has `combine(1...n)(implicit composition)` and `combineWith(1...n)(combinator)` methods, EventStream instances have the same plus also `withCurrentValueOf(1...n)(implicit composition)` and `sample(1...n)` in a similar manner. No `*with` versions of those, just use  `map` / `mapN` for that. Need to keep the number of methods reasonable.

Speaking of, maybe we should generate `map{1..N}(...)` and `filter{1..N}(...)` methods too. Kinda lame that they need a number in the name, but IIRC we can't avoid that due to either type erasure or type inference issues (don't remember exactly).

All of the above is also done for signals btw.

@yurique Am I using `Composition` as intended in [CombinableEventStream](https://github.com/raquo/Airstream/pull/51/files#diff-f2f41b9ac311be1fdd1de8e128a817c55802ef80eb5f5323a2768fe089c101ac)? I was hoping to avoid the (a, v1, v2) => (a, (v1, v2)) lambdas, but I don't think it's doable.

yurique:
@raquo this looks great!

I've got a few errands to run :) – I'll look closely a bit later. On the first glance – it looks like avoiding those lambdas is not possible in this case indeed,  as `compose` takes two parts to compose, and here you first need to build the second tuple.

raquo:
Cool, no rush, I have a mountain of other issues / PRs to sort through
anyway!

--Nikita

yurique:
Just a thought: I can see myself using primarily `Signal`/`EventStream` `.seq`/`.combine` most of the time. So doing extra work to maintain the extra `combine(s1, s2,  ...)`, `sample(s1, s2, ...)` and `withCurrentValueOf(s1, s2, ...)` might not be necessary.

```scala
    import com.raquo.airstream.signal.Signal
    import com.raquo.airstream.signal.Val
    import com.raquo.airstream.combine.generated.{ CombinableEventStream, CombinableSignal }

    class A
    class B
    class C
    class D
    class E

    val stream: EventStream[A] = EventStream.empty
    val signal: Signal[A] = Val(new A)

    val otherStream1: EventStream[(B, C)] = EventStream.empty
    val otherStream2: EventStream[(D, E)] = EventStream.empty

    val signal1: Signal[(B, C)] = Val((new B, new C))
    val signal2: Signal[(D, E)] = Val((new D, new E))


    val _: EventStream[(A, (B, C), (D, E))] = new CombinableEventStream(stream).combine(otherStream1, otherStream2)
    // is equivalent to:
    val _: EventStream[(A, (B, C), (D, E))] = EventStream.combine(stream, otherStream1, otherStream2)
    // the user most likely wants this instead in this case:
    val _: EventStream[(A, B, C, D, E)] = stream.combine(otherStream1).combine(otherStream2)

    val _: EventStream[(A, (B, C), (D, E))] = new CombinableEventStream(stream).withCurrentValueOf(signal1, signal2)
    // is equivalent to:
    val _: EventStream[(A, (B, C), (D, E))] = stream.withCurrentValueOf(Signal.combine(signal1, signal2))
    // the user most likely wants this instead in this case:
    val _: EventStream[(A, B, C, D, E)] = stream.withCurrentValueOf(signal1).withCurrentValueOf(signal2)

    val _: Signal[(A, (B, C), (D, E))] = new CombinableSignal(signal).combine(signal1, signal2)
    // is equivalent to:
    val _: Signal[(A, (B, C), (D, E))] = Signal.combine(signal, signal1, signal2)
    // the user most likely wants this instead in this case:
    val _: Signal[(A, B, C, D, E)] = signal.combine(signal1).combine(signal2)
```


Also there's a gotcha – currently, the extra combine methods provided by `implicit def toCombineOpsStream` and `implicit def toCombinableSignal` are not available, because `EventStream` and `Signal` have a `combine` method defined on them, so the compiler will not try to find an implicit conversion and just complains:

```
too many arguments (found 2, expected 1) for method combine: (otherSignal: com.raquo.airstream.signal.Signal[B])(implicit composition: app.tulz.tuplez.Composition[A,B]): com.raquo.airstream.signal.Signal[composition.Composed]
```

For this to work, either:
* all combine overloads (including the one currently defined in `Signal` / `EventStream`) should be defined in the `CombinableEventStream`/`CombinableSignal`
* or the `Signal`/`EventStream` should **extend** `CombinableEventStream`/`CombinableSignal`

### combineSeq

We should probably name it `Signal.sequence(...)` / `EventStream.sequence(...)` as this is the standard name for a transformation from `Seq[F[A]]` into `F[Seq[A]]`.

Futures:

```scala
    val seqOfFuture: Seq[Future[Int]] = Seq(Future.successful(1), Future.successful(2))
    val _          : Future[Seq[Int]] = Future.sequence(seqOfFuture)
```

Cats effect:
```scala
import cats.effect.IO
import cats.syntax.all._

val io1: IO[Int] = IO.pure(1)
val io2: IO[Int] = IO.pure(2)
val io3: IO[Int] = IO.pure(3)
val io4: IO[Int] = IO.pure(4)

val seqOfIo: Seq[IO[Int]] = Seq(io1, io2, io3, io4)
val ioOfSeq: IO[Seq[Int]] = seqOfIo.sequence
```

https://scastie.scala-lang.org/Va4D3IHaRpuZx9ioXuKr1w


And for Airstream it would be:
```scala
    val seqOfSignal: Seq[Signal[Int]] = Seq(Val(1), Val(2), Val(3))
    val _:           Signal[Seq[Int]] = Signal.sequence(seqOfSignal)

    val seqOfStream: Seq[EventStream[Int]] = Seq(EventStream.empty, EventStream.empty, EventStream.empty)
    val _:           EventStream[Seq[Int]] = EventStream.sequence(seqOfStream)
```

yurique:
Also (and I'm not suggesting actually doing it :) ), there's the `tuplez-apply` libs which should allow doing this:

```scala
val signal: Signal[A, B, C, D, E] = ???
signal.map( (a, b, c, d, e) =>  ??? ) // not case (a, b, c, d, e)
```

Not sure if anyone would prefer this, but it looks nice in frontroute.

yurique:
```scala
val seqOfIo: Seq[IO[Int]] = Seq(io1, io2, io3, io4)
val ioOfSeq: IO[Seq[Int]] = seqOfIo.sequence
```

I actually like this one, too. We might also support this syntax (though this is probably an overkill and definitely less discoverable vs `Signa.sequence`/`EventStream.sequence`):

```scala
    val seqOfSignal: Seq[Signal[Int]] = Seq(Val(1), Val(2), Val(3))
    val _:           Signal[Seq[Int]] = seqOfSignal.sequence

    val seqOfStream: Seq[EventStream[Int]] = Seq(EventStream.empty, EventStream.empty, EventStream.empty)
    val _:           EventStream[Seq[Int]] = seqOfStream.sequence
```

yurique:
```scala
signal.map( (a, b, c, d, e) =>  ??? ) // not case (a, b, c, d, e)
```

this should probably be `mapN`:
```scala
signal.mapN( (a, b, c, d, e) =>  ??? )
```

raquo:
Thanks for the review @yurique, I addressed the issues.

- Using `sequence` name now (but I kept the `combineSeq` alias for some autocomplete visibility and consistency with mergeSeq) 
- Fixed the compilation issue with implicits. Made all combine-related instance methods with the implicit. I figured we're already doing the same with `split` and no one complained about it so far
- I kept the N-versions of withCurrentValueOf and sample. You're right that there are equivalent combine incantations that will achieve the same result, but the dedicated operators that I made have better performance (less combine observables, less pending / syncing logic).
- The reason I want instance methods of combine / combineWith is because almost all other operators are available on instances (except merge / mergeSeq), and also because in my Laminar app I sometimes use a pattern like `inContext { _.events(onClick).combineWith(...) --> obs }`, and without that instance `combineWith` method it would need to be quite a bit more verbose, having to spell out `thisNode` and stuff, with proper formatting would probably take several lines.
- At any rate, there is no runtime performance penalty for the extra instance methods given how Scala / Scala.js encodes implicit value class method calls. Biggest annoyance is having to pollute the autocomplete results with several versions of combine*, but that's pretty much unavoidable whenever you start typing that method name.
- I decided to reduce all the generation and tupling to a maximum of Tuple9 (prev was Tuple10). It's an arbitrary limit either way, just wanted to reduce the amount of cruft and keep stuff in single digits. You don't need to update tuplez, it's fine as is.
- Regarding implicit extension methods on foreign types like Seq[EventStream].sequence or Future.toSignal, we don't have any of those yet, and I'm also concerned about discoverability / redundancy.

yurique:
Looks great! :tada:  I'll give it a more in-depth look a bit later today.

A small question/suggestion:

Should `combine`s be `@inline` here?

```scala
  def combine[T1, T2, T3, T4](
    s1: Signal[T1], s2: Signal[T2], s3: Signal[T3], s4: Signal[T4]
  ): Signal[(T1, T2, T3, T4)] = {
    combineWith(s1, s2, s3, s4)(Tuple4.apply[T1, T2, T3, T4])
  }

  def combineWith[T1, T2, T3, T4, Out](
    s1: Signal[T1], s2: Signal[T2], s3: Signal[T3], s4: Signal[T4]
  )(
    combinator: (T1, T2, T3, T4) => Out
  ): Signal[Out] = ... 
```

yurique:
@raquo this needs to be removed (publish fails otherwise as the files are generated into the unmanaged sources dir):

```scala
mappings in (Compile, packageSrc) ++= {
  val base  = (sourceManaged in Compile).value
  val files = (managedSources in Compile).value
  files.map { f =>
    (f, f.relativeTo(base / "scala").get.getPath)
  }
}
```

yurique:
@raquo do you mind updating the tuplez version to `0.3.1`? 

The tuple modules (full, basic, etc) have not been changed, but I've fixed a bug in `tuplez-apply` and bumped the version for all of the modules.

yurique:
I've built next-0.12 snapshots of Laminar and Airstream locally, updated laminext and a project I'm building with Laminar to use them – everything works like a charm 👍 

The only thing I'm concerned about is `StaticSignalCombineOps` being limited to just 9 signals with no way to expand this.

(I'm using these to combine signals holding input values for a form into one signal and I might end up having a form with > 9 inputs at some point 🤔 )

Maybe we could have `Airstream` and `Airstream-full` ? 👼  😄

yurique:
I mean, there are ways to expand it, but not with the same name.

What if instead of 

```scala
object Signal extends StaticSignalCombineOps {
```

we had 

```scala
object Signal {

  implicit def signalCompanionCombineSyntax(s: Signal.type): StaticSignalCombineOps.type = StaticSignalCombineOps

}
```

and

```scala
object StaticSignalCombineOps {} 
```

instead of 

```scala
trait StaticSignalCombineOps {} 
```
?

This way it would be possible to have another `implicit def something(s: Signal.type)` which would provide other overloads.

And, of course, I can have the extra combines defined in a different object.

So there are definitely some escape hatches. But I'd rather have my autocomplete spammed with another dozen of `combine` options :)

raquo:
Very good points @yurique, thanks, I've now changed static methods from inheritance to implicits to allow easy extensions by end users.

This should provide enough flexibility so that I don't want to mess with airstream packages and publishing (normal vs `full`).

yurique:
Thank you, @raquo, this is super exciting! (I tried rebuilding all I have with a new snapshot – it works!)

PR #54, opened 2020-12-27
Web platform streams

raquo:
Hey @ajaychandran thanks for your #53 PR!

I applied some formatting to your commit, and couldn't force-push it to your branch so I opened this PR. There might be some github PR flow that I'm missing... Anyway, 31493bf is your original commit patched with my formatting changes. I just added you as collaborator, so next time feel free to just create a branch straight on raquo/Airstream.

If this looks good to you, I'll merge.

Closes #53


README.md
```diff
@@ -563,12 +566,54 @@ Remember that this atomicity guarantee only applies to failures which would have
...
+AjaxEventStream
+  .get("/api/kittens") // EventStream[dom.XMLHttpRequest]
+  .map(req => req.responseText) // EventStream[String]
+```
+
+Methods for POST, PUT, PATCH, and DELETE are also available.
+
+The request is made every time the stream is started. If the stream is stopped while the request is pending, the request will not be cancelled, but its result will be discarded.
+
+The implementation follows that of `org.scalajs.dom.ext.ajax.apply`, but is adjusted slightly to be better behaved in Airstream.
```
ajaychandran, line 584: Change to `org.scalajs.dom.ext.Ajax.apply`

ajaychandran:
LGTM.

Also, it would be nice to standardize formatting using the [sbt-scalafmt](https://github.com/scalameta/sbt-scalafmt) plugin.

src/main/scala/com/raquo/airstream/web/AjaxEventStream.scala
```diff
@@ -0,0 +1,162 @@
...
+  * On restart, a new request is performed and the subsequent success/failure is propagated downstream.
+  */
+class AjaxEventStream(
+  method: String,
+  url: String,
+  data: dom.ext.Ajax.InputData,
+  timeout: Int,
+  headers: Map[String, String],
+  withCredentials: Boolean,
+  responseType: String
```
yurique, line 23: Can we also have a progress observer here?

Something like this:
 
```scala
maybeProgressObserver: Option[Observer[ProgressEvent]]
```

and below
```scala
    maybeProgressObserver.foreach { progressObserver =>
      req.upload.onprogress = p => progressObserver.onNext(p)
    }
```

And add this param to the `apply`/`post`/`put` etc.

There are other events as well:

```scala

  /**
   * A function that is called periodically with information about the progress of the
   * request.
   *
   * MDN
   */
  var onprogress: js.Function1[ProgressEvent, _] = js.native

  /**
   * The function to call when a request encounters an error.
   *
   * MDN
   */
  var onerror: js.Function1[ErrorEvent, _] = js.native

  /**
   * The function to call when an HTTP request returns after successfully loading
   * content.
   *
   * MDN
   */
  var onload: js.Function1[js.Any, _] = js.native

  /**
   * A function that is called if the event times out; this only happens if a timeout has
   * been previously established by setting the value of the XMLHttpRequest object's
   * timeout attribute.
   *
   * MDN
   */
  var ontimeout: js.Function1[js.Any, _] = js.native

  /**
   * The function to call when a request is aborted.
   *
   * MDN
   */
  var onabort: js.Function1[js.Any, _] = js.native

  /**
   * A function that gets called when the HTTP request first begins loading data.
   *
   * MDN
   */
  var onloadstart: js.Function1[js.Any, _] = js.native

  /**
   * A function that is called when the load is completed, even if the request failed.
   *
   * MDN
   */
  var onloadend: js.Function1[ProgressEvent, _] = js.native
```

But so far I've only ever needed the `onprogress`.

raquo:
@ajaychandran Sadly scalafmt doesn't work for me. Its output is never what I want no matter how I configure it, especially its insistence on handling newlines instead of just letting the developer put whatever newslines wherever they want. The rules they have are not nearly flexible enough. I can tolerate scalafmt at work where there's a lot of collaboration going on, but on my own projects it's just easier for me to auto-format with intellij when a PR style doesn't match the codebase. Yes, it's stupid but it works. I wish Scala had something like `eslint`, which knows to get out of your way, but as far as I know there is no such thing.

yurique:
@raquo there's this setting for scalafmt:

```
optIn.breaksInsideChains = true
```

it allows doing this:
```
  something.f()
    .f2() // it won't try collapsing these into one line
    .f3()
```

after I enabled this, I could live with what scalafmt does, but yeah, a few things remain that I don't like (but accepted :) ) — the biggest is it's inconsistency when formatting for blocks with `<-` and `=` like here:

```scala
  for {
    a    <- q()
    abcd <- p()
    something <- f {
                   someCode()
                   here
                 }
    dcba <- o()
    b = g()
  } yield ???
```

I want all `<-` and `=` to be alligned, but no.

src/main/scala/com/raquo/airstream/web/AjaxEventStream.scala
raquo: I'm working on this, it got deeper (but will end up better).

ajaychandran:
The consistency of formatting has significantly improved in recent versions. You may want to give it another try.

Source [line breaks](https://scalameta.org/scalafmt/docs/configuration.html#newlinessource) can be retained with
```
newlines.source=keep
```

Also, formatting can be [turned off](https://scalameta.org/scalafmt/docs/configuration.html#disabling-or-customizing-formatting) with
```
// format: off
```

Not having standard formatting is a real pain point.

raquo:
> Source line breaks can be retained with

> > newlines.source=keep

Thanks @ajaychandran that would actually be very nice if it works as expected. Looks like they added this option earlier this year. I'll give it a shot, but I have to prioritize clearing up all pending PRs first.

raquo:
@yurique I've added support for progress events to AjaxEventStream. That prompted me to reevaluate my desire to keep the ajax integration minimal / unopinionated, because I realized how hard it would be for end users to add custom event listeners to the stream to e.g. distinguish timeouts / aborts / normal errors, so I decided to make the wrapper more opinionated but also more useful in Airstream. It's still pretty plain, but errors are now reported in a structured way.

Also I came up with a new, easier way to create custom streams and signals from external sources (be it ZIO, ajax, futures, etc.). I didn't actually end up using it for Ajax because I wanted that custom `val responseEvents`. I haven't actually documented this yet, but the API is available via `CustomEventSource` and `CustomSignalSource`. The basic idea is to take away some of the foot guns that are available when extending EventStream or Signal, and to provide a short guide for using the restricted CustomSource API. (that one is todo)

Also added a few nice things like `val (callback, stream) = EventStream.withCallback[Foo]` to simplify canonical use cases of `EventBus`, integrations with JS code, etc. (`stream` emits the events fired into `callback`)

raquo:
FYI I haven't actually tested the new ajax stream yet. Not sure if adding live network tests is a good idea, but I'd love to have a page like http://plnkr.co/edit/ycQbBr0vr7ceUP2p6PHy?preview (but using AjaxEventStream) on Laminar website.

ajaychandran:
Constructors such as `EventStream.fromSeq` require an `emitOnce` parameter.

Could such constructors be replaced with
```scala
object EventStream {

  def fromSeq[A](xs: Seq[A]): EventStream[A] = ???
  def repeatSeq[A](xs: Seq[A]): EventStream[A] = ???

  // or the names could be emitSeqOnce/emitSeq
}
```

This would improve ergonomics at call site.

```scala
// current
// standard code conventions require naming boolean parameters
val s1 = EventStream.fromSeq(List(1), emitOnce = true)
val s2 = EventStream.fromSeq(List(1), emitOnce = false)

// proposed
// smaller code, intent is much more clear
val s3 = EventStream.fromSeq(List(1))
val s4 = EventStream.repeatSeq(List(1))
```

raquo:
I made a simple tester for Ajax requests here https://github.com/raquo/laminar-examples/pull/6 and fixed some bugs that I found that way. I only tested simple GET requests so far.

To keep my sanity, I'm gonna merge this PR into `next-0.12` now, it's grown too much beyond Ajax. That doesn't mean the work on Ajax is finished, further improvements should just happen in new PRs. The question of get/post/etc helper methods API is still open if anyone still wants to weigh in.

Issue #43, opened 2020-11-22
Re-evaluate signal's current value when restarting signal


raquo:
```scala
val state =  Var(0)
val x10signal = state.signal.map(_ * 10)
val x100signal = x10signal.map(_ * 10)
```

Currently, when you **first start** `x100signal`, it initializes its initial value (not a typo – signals' initial values are lazy) by pulling the current value of `x10signal`, which initializes the initial value of `x10signal` by pulling `state.signal`'s current value. However, if afterwards you **stop** `x100signal`, then after some time **start** it again, this pulling-current-value chain will not happen, `x100signal` will simply start listening to new `x10signal` updates from this point on. This means that if `state` was updated while `x100signal` was stopped, `x100signal` will not receive this update when it's started again.

This is documented behavior – observables are lazy, so you shouldn't stop the ones you care about – but it would be more useful if `x100signal` could recursively pull its parent's current value not just when initializing the signal's initial value, but every time it's started.

This would have convenient practical effects in Laminar when re-mounting unmounted elements – those would be brought up to date when they're re-mounted.

However, very importantly, this measure can't cancel out laziness per se, and specifically this can not possibly replay the EventStream events that fired while the signal was stopped, so we will not be able to pull an updated current value for signals that depend on event streams, such as `stream.foldLeft(...)` or `stream.startWith(...)`.

This caveat will make undesired behavior more rare, but more specialized and harder to debug. However, it is probably a good tradeoff, since it should be more obvious that any events you've missed will not be re-emitted, than how your signal won't pull a parent's current value when it's reactivated.

[gitter context](https://gitter.im/Laminar_/Lobby?at=5fb9e19dbba6a3778b75267e)

PR #60, opened 2020-12-30
Added websocket event sources

ajaychandran:
I have tested the implementation for binary and text data. Included support for `dom.Blob` data type, not sure how to test it though.

This [example](https://gist.github.com/ajaychandran/c7ba1ddd423aed826e03d7f1edd92890) can be added to the Laminar project once the PR is merged and released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -587,41 +587,69 @@ The implementation follows that of `org.scalajs.dom.ext.ajax.apply`, but is adju
 
 ### Websockets
 
-Airstream supports uni-directional and bi-directional websockets.
+Airstream supports unidirectional and bidirectional websockets.
 
+#### Absolute URL is required
+```scala
+import com.raquo.airstream.web.websocketUrl
+
+val url: String = websocketUrl("relative/url")
+```
+
+#### Unidirectional websocket stream
 ```scala
 import com.raquo.airstream.eventstream.EventStream
 import com.raquo.airstream.web.WebSocketEventStream
 import org.scalajs.dom
 
-import scala.scalajs.js.typedarray.ArrayBuffer
+// builder for creating unidirectional stream
+val builder = WebSocketEventStream("absolute/url")
 
-// absolute URL is required
-// use com.raquo.airstream.web.websocketUrl to construct an absolute URL from a relative one
-val url: String = ???
+// raw websocket messages
+val raw: EventStream[dom.MessageEvent] = builder.raw
 
-// uni-directional, server -> client
-val s1: EventStream[dom.MessageEvent] = WebSocketEventStream(url)
+// extract and cast dom.MessageEvent.data
+val data: EventStream[String] = builder.data[String]
 
-// bi-directional, transmit text from client -> server
-val src2: EventStream[String] = ???
-val s2: EventStream[dom.MessageEvent] = WebSocketEventStream(url, src2)
-
-// bi-directional, transmit binary from client -> server
-val src3: EventStream[ArrayBuffer] = ???
-val s3: EventStream[dom.MessageEvent] = WebSocketEventStream(url, src3)
-
-// bi-directional, transmit blob from client -> server
-val src4: EventStream[dom.Blob] = ???
-val s4: EventStream[dom.MessageEvent] = WebSocketEventStream(url, src4)
+// alias for the common usecase (data[String])
+val text: EventStream[String] = builder.text
 ```
 
-The behavior of the returned stream is as follows:
- - A new connection is established when this stream is started.
- - Upstream messages, if any, are transmitted on this connection.
- - Server messages are propagated downstream.
- - Connection termination, not initiated by this stream, is propagated downstream as an error.
- - The connection is closed when this stream is stopped.
+#### Bidirectional websocket stream
+Usage:
+```scala
+import com.raquo.airstream.eventstream.EventStream
+import com.raquo.airstream.web.WebSocketEventStream
+import org.scalajs.dom
+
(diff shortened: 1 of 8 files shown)


ajaychandran:
Looks like I started with an older version of the target branch. Let me know if you want me to resubmit the PR.

ajaychandran:
@raquo Please review.

yurique:
I have a WebSocket API in laminext, and I took a different approach there (I just added support for blob/arrays inspired by this PR :) ).

The usage looks like this:

```scala
      val ws = websocket.forUrl("wss://echo.websocket.org").string
// or
      websocket.forUrl("wss://echo.websocket.org").text(decode, encode)
      websocket.forUrl("wss://echo.websocket.org").blob
      websocket.forUrl("wss://echo.websocket.org").arraybuffer
// or      
      websocket.forUrl("wss://echo.websocket.org").receiveString.sendArray
      websocket.forUrl("wss://echo.websocket.org").receiveBlob.sendText(decode, encode)
// and all other combinations 
```

and then there's a bunch of binders + a stream and an observer provided:

simplest:
```scala
val outgoingEvents: Observable[String] = ???

div(
  ws.connect, //only needed when no other binders are used, like in this example
  child.text <-- ws.received.stream,
  outgoingEvents --> ws.sendObserver
)
```

other binders:
```scala
div(
  ws.send <-- outgoingEvents
)
```

```scala
div(
  ws.connected --> connectedObserver,  // or () => Unit, as usual
  ws.closed --> closedObserver,  
  ws.error --> errorObserver,  
  ws.received --> receivedObserver
)
```

It buffers the outgoing messages until connected (configurable). A couple of things remain to be done, like the re-connection logic.

The working example can look like this:

```scala
val ws = websocket.forUrl("wss://echo.websocket.org").string
val inputElement = input()
div(
  div(inputElement),
  div(  
    button(
      "send",
      onClick.mapTo(inputElement.ref.value) --> ws.sendObserver
    )
  ),
  ws.connect,
  div(text <-- ws.received)
)
```

yurique:
It's a quite old piece of code, though. Now I'm looking at it and wondering if I could ditch half of it (together with some extra complexity)

yurique:
after some cleanup my implementation looks like this: https://gist.github.com/yurique/baae21c9eea620033e0fc2cf5f370701

+ a couple of files with builders

raquo:
I reviewed this branch, Iurii's gist, and my own private websocket implementation (my own stuff is not very useful because it's too tied to my personal code and the libraries that I use), and after all, I don't think adding websockets to Airstream is the right thing to do now.

The amount of complexity required to make a reasonably safe, resilient and pleasant API is very significant, and I don't have the capacity right now to get to the bottom of it and make the right call. Things that need to be designed, at the very least, are:
* handling errors and disconnects, and maybe automatic reconnects
* queing of pending client-to-server messages while the the connection was not yet open or while it was disconnected
* library needs to be aware that users will most likely use a codec for both inbound and outbound messages, and so should probably support such config

It's not clear to me how the API should be designed in terms of stream lifecycle either.

Iurii's implementation depends on Laminar but provides a seemingly more convenient API. I honestly don't know if that's the way to go or not. I think we'll need to see Iurii's laminext released, presumably with his websockets implementation, and see how that fares, what kind of feedback we get about such an approach to websockets.

Websockets are quite far from Airstream's core functionality, so I'm ok having examples / gists / third party libraries cover that for now.

At any rate, this turned out to be too big of an issue to include in v0.12.0. I'm going to fix a bunch more small stuff and wrap up that release now.

yurique:
This is a very exciting release!

I'll do my best to release laminext asap after 0.12 is out. Almost everything is prepared, just a couple of modules left to document.

ngbinh:
should this be included as an extension module of Airstream instead? Of course, post `0.12`?

raquo:
Well, that's what Iurii's laminext project will be, right, an extension
library for Laminar & Airstream. Ajay's implementation can be published
similarly, it doesn't have to be part of Airstream core.

The benefit of having third parties publish these extensions is that I
don't have to participate in the design or maintain the code afterwards,
and you can have several alternatives on equal standing.

I mean eventually it would be nice to have a canonical websockets wrapper
in Airstream itself, but it's a lot of work to do it the way I think it
should be done (I didn't realize this initially, or I would have said it
upfront), and the cost / benefit ratio isn't there yet. In my view issues
like https://github.com/raquo/Airstream/issues/43 are more pressing, and
that isn't an easy problem either, that will take a lot of dev time too.

—
Best regards,
Nikita

ajaychandran:
I have redesigned the API to support connection management.
```scala
// a duplex connection
// (Observer[Boolean], Observer[String[, EventStream[String])
val (control, transmit, receive) = WebSocketEventStream.text.open("absolute/url")
```
- a connection is established when `receive` is started
- sending `true` to `control` will reopen a connection (provided receive is started)
- sending `false` to `control` will close the connection
-  the connection is closed when `receive` is stopped
-  `control` can be used in conjunction with a `socketErrorObserver` (optional constructor parameter) to implement an automatic retry policy.

> At any rate, this turned out to be too big of an issue to include in v0.12.0. I'm going to fix a bunch more small stuff and wrap up that release now.

I agree.
It would be nice to get some feedback from developers using any of the implementations presented here.

doofin:
Hi @ajaychandran , as people mentioned before laminext https://github.com/tulz-app/laminext is the right place to put this, so Airstream won't merge this. let's discuss if we can contrib in laminext

PR #64, opened 2021-01-04
Fetch event stream

yurique:
I tried implementing the `FetchEventStream` for requests using `fetch` which is supposed to be a "modern replacement for `XMLHttpRequest`" (unfortunately, fetch still doesn't allow to watch the upload progress, but this is an edge case).

In overall, `fetch`'s API is much better – base on promises.

The API here looks like this:

```scala
Fetch.get("https://httpbin.org/get").text // EventStream[FetchResponse[String]] 
Fetch.get("https://httpbin.org/get").json // EventStream[FetchResponse[js.Any]]
Fetch.get("https://httpbin.org/get").blob // EventStream[FetchResponse[Blob]]
Fetch.get("https://httpbin.org/get").arrayBuffer // EventStream[FetchResponse[String]]
Fetch.get("https://httpbin.org/get").raw // EventStream[FetchResponse[Response]]
Fetch.get("https://httpbin.org/get").readableStream // EventStream[FetchResponse[ReadableStream[Uint8Array]]]
```

```scala
  final case class FetchResponse[A](
    ok: Boolean,
    status: Int,
    statusText: ByteString,
    headers: Headers,
    `type`: ResponseType,
    data: A,
    url: String,
  )
```

`Fetch` also contains builders for all http methods, and the builder can be configured before creating an event stream:

```scala
Fetch.post("https://httpbin.org/get", headers = Map("a" -> "b"), body = ???)
  .referrer(???)
  .referrerPolicy(???)
  .mode(???)
  .credentials(???)
  .cache(???)
  .integrity(???)
  .keepalive(???)
  .timeout(???)
  .text

Fetch.post("https://httpbin.org/get", headers = Map("a" -> "b"), body = ???)
  .configure( ... )
  .text
```

Timeout is implemented within the `FetchEventStream`.

In the `onStop` and when timing out, `FetchEventStream` aborts the actual request (using the `AbortController`).


yurique:
upload streaming (and progress tracking) is coming someday: https://www.chromestatus.com/features/5274139738767360

yurique:
download progress can be implemented: https://javascript.info/fetch-progress

raquo:
Thanks!

I haven't used Fetch myself directly yet, is there any advantage to having both XHR and Fetch implementations?

API-wise, I would like to keep Ajax, Fetch, and ideally websockets have a similar feel. We aren't really using the builder pattern anywhere at the moment, so I'm not a fan of using it here if the same can be achieved by optional parameters like in AjaxEventStream.

Note, I haven't had the chance to review the websockets stuff yet, I will need to look at that too, to see what kind of API makes sense there.

yurique:
> is there any advantage to having both XHR and Fetch implementations

There's just one: XHR allows to observe the upload progress, fetch – doesn't (someday it will, but it's unclear). 
Otherwise, fetch is a much better designed API. And there are some differences, like fetch always succeeds (regardless of the response status code). 

> I would like Ajax, Fetch, and ideally websockets to have a similar feel

Totally 👍 

> if the same can be achieved by optional parameters like in AjaxEventStream.

I started out with copying the `AjaxEventStream` and initially had the same pattern with get/post/etc functions accepting optional params and passing them to the `FetchEventStream` constructor, and I don't mind having it like that.

The tricky part is getting the response:

fetch returns a `Promise[Response]`, and this `Response` object contains status code, headers, etc but not the response body. For the response body it provides methods like `.text()`, `.json()`, `.blob()`, etc, which, in turn, also return a Promise (except the `.body` which returns a `ReadableStream`, which nonetheless also needs to be processed before you get the response itself). 

So just having just an `EventStream[Response]` is not very convenient in most cases. Thus, I ended up with `.get/.post./etc -> .text/.json/.blob/etc`. Having `getText/getBlob/.../postText/postBlob/...` doesn't seem feasible.

But I might be missing some obvious solution here, so I'm open-minded :)

raquo:
> Otherwise, fetch is a much better designed API. And there are some differences, like fetch always succeeds (regardless of the response status code).

To clarify my concern... I _can_ see why Fetch would be seen as a better designed API when compared to XHR in _plain JS_. However, what is the purpose for making a wrapper for it _in Airstream_? If XHR and Fetch are close enough substitutes for each other, our Airstream Ajax API will erase any differences between them, e.g. between the callback vs promise style, leaving only the differences in functionality.

But is there something _important_ in the Fetch API that's missing from XHR? Because if there is nothing substantial like that, then the AjaxEventStream that we have already covers the Ajax functionality well enough, and if that's the case then I would prefer for a Fetch interface to be part of laminext rather than airstream, especially seeing that wrangling Fetch API into a pleasant Airstream interface might not be trivial.

yurique:
There's a couple of differences. First, dealing with `fetch`'s `Response` (or any of the body-"getters") feels more natural than having to think about the `XMLHttpRequest` and remembering how it works.

And another one, which I find the most annoying, is that `XMLHttpRequest` fails on any response status except 2xx (I prefer to see if it was a 400 or a 500 and deal with it in a normal way). Not that this is not possible with `XMLHttpRequest`, but it's definitely more messed up.

I can definitely maintain it in laminext – I was planning to put it there (probably as a separate module containing only it),  wanted to share it here just in case :)

yurique:
I just checked to make sure – currently, `AjaxEventStream` also fires an error when the response is non-2xx.

raquo:
@yurique I just added some customization and error structure for http codes: https://github.com/raquo/Airstream/pull/48/commits/af90ac6a18b32939d23134e20e2c9f91124a0a2e Regardless of Fetch, that's a good idea.

The "only 2xx" response thing is not built-in to XHR, it was our own filter (that I now made customizable).

One really annoying thing about XHR is I can't figure out how to get the details of the network errors. Is Fetch any better in that regard?

yurique:
Network errors are related to CORS, so the browser just doesn't expose anything to JS in most (all?) cases because of that. I had the same thing with websockets – trying to figure out how to get the error message (when the connection fails, etc), turned out the browser just doesn't give up any information at all.

raquo:
What do you know, I finally got to this. @yurique's Fetch implementation has been living in Laminext all this time.

Now, despite [memetic advice to the contrary](https://media.makeameme.org/created/stop-trying-to-xjvtkv.jpg), I'm adding FetchStream to Airstream in the next-0.15 branch ([commit](https://github.com/raquo/Airstream/commit/86bebe33a5fe9cf3db721f6e94f92b1f1257376e)). I made a slightly different API though. Usage of my FetchStream is like this:

```scala
FetchStream.get(
  "https://example.com",
  _.headers(
    "key1" -> "value1",
    "key2" -> "value2"
  ),
  _.mode(_.cors),
  _.redirect(_.follow),
  _.abortStream(anyEventStream)
) // returns EventStream[String] of `responseBody`

FetchStream.raw.post("https://example.com/submit", ...) // returns EventStream[dom.Response]

FetchStream.withCodec(encodeBody, decodeResponseIntoStream)(
  _.POST,
  "https://example.com",
  _.headersAppend("key" -> "value"),
  _.body(bodyToBeEncoded), // request body
  _.redirect(_.follow),
  _.referrerPolicy(_.`no-referrer`)
) // returns EventStream[YourDecodedResponse]

// etc.
```

PR #69, opened 2021-01-27
Make a Signal forget its internal memory when stopped.

### LINKED ISSUE #43: Re-evaluate signal's current value when restarting signal

vic:
Our team came across the issue described in #43. 

We have a Laminar element rendering a value mapped from a Signal: `val fooView = div(child.text <-- fooSignal.map(_.toString))`. However if `fooView` was unmounted and meanwhile the `fooSignal` received updates, when `fooView` is re-mounted it displayed out-of-date values (the latest value it was able to see before being unmounted) and not the current value from `fooSignal`.

I understand this is the expected behaviour, since Airstream signals are lazy which is totally OK!.
However people in our team were expecting that once a Signal gets re-started it should ask their parent
signal for the newest value (just like when first activated).

This branch changes `Signal.onStop` to make it forget its internal memory when all observers have been removed by just reverting `maybeLastSeenCurrentValue` to its original undefined state, causing it
to re-compute the "initialValue" again if the Signal is re-started.

This way, elements rendering values from dependent signals (eg, MapSignal) correctly show the latest
value when re-mounted.

We are currently using this branch from jitpack and all seems to work as expected. However I was not able to run all tests (since I'll have to install yarn and probably other dependencies), but I would like to know if this PR's approach is correct.

Fixes #43

### DIFF
--- a/src/main/scala/com/raquo/airstream/signal/Signal.scala
+++ b/src/main/scala/com/raquo/airstream/signal/Signal.scala
@@ -137,6 +137,11 @@ trait Signal[+A] extends Observable[A] {
     super.onStart()
   }
 
+  override protected[this] def onStop(): Unit = {
+    super.onStop()
+    maybeLastSeenCurrentValue = js.undefined
+  }
+
   // @TODO[API] Use pattern match instead when isInstanceOf performance is fixed: https://github.com/scala-js/scala-js/issues/2066
   override protected def onAddedExternalObserver(observer: Observer[A]): Unit = {
     super.onAddedExternalObserver(observer)


raquo:
Thanks for the PR!

I haven't had the time to analyze the #43 issue in depth yet (it's not going to make it into 0.12.0) so I can't say for sure if this solution is on the right track. I have some doubt, because conceptually, the Signal shouldn't forget its state when it's stopped, it should instead re-evaluate its state when it's (re)started.

What I mean is, the signal's current value is _always_ available via `tryNow()`. If we clear `maybeLastSeenCurrentValue` when the signal is stopped, then if anything accesses the signal's `tryNow()` while the signal is stopped, the signal's current value will be re-evaluated, which we don't want, we want to get the old, potentially stale, value instead, because signals shouldn't change their value while they're stopped.

On the other hand, a signal's `tryNow()` method is not public. Users will not be accessing it directly. I'm not sure if Airstream can guarantee that it doesn't call `tryNow()` internally while the signal is stopped. That might well be true, but only accidentally, not by design. And since users can extend `EventStream` and `Signal` classes to write custom logic that has access to the signal's `tryNow` method by such means, users need to be made aware that they shouldn't call tryNow() on stopped signals at the very least.

I'll need to look more into this at some point after 0.12.0 is released.

raquo:
I've implemented a proper solution to this issue in the `next-0.15` branch. See #43 #93

PR #71, opened 2021-03-07
Add Observer#contramapOpt

jatcwang:
contramapOpt will not call original observer's onNext if the provided transformation returns a None.

### DIFF
--- a/src/test/scala/com/raquo/airstream/errors/ObserverErrorSpec.scala
+++ b/src/test/scala/com/raquo/airstream/errors/ObserverErrorSpec.scala
@@ -1,14 +1,14 @@
 package com.raquo.airstream.errors
 
-import com.raquo.airstream.UnitSpec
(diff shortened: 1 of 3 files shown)


raquo:
Nice, thanks!

src/test/scala/com/raquo/airstream/errors/ObserverErrorSpec.scala
```diff
@@ -301,8 +301,6 @@ class ObserverErrorSpec extends UnitSpec with BeforeAndAfter {
 
     // -- Now check the contramapped observer behaviour
 
-    lowObs.onNext(1)
```
raquo, line 304: Couple things I fixed in tests:
- This was obviously removed accidentally
- Scala 2.12 does not recognize `a => b` as a partial function, it needs `{ case a => b}`

jatcwang: Whoopsie my bad. Thanks :)

PR #74, opened 2021-03-21
Scala 3 support

yurique:


src/main/scala/com/raquo/airstream/core/BaseObservable.scala
```diff
@@ -72,13 +71,15 @@ trait BaseObservable[+Self[+_] <: Observable[_], +A] extends Source[A] with Name
     this match {
       case s: Signal[A @unchecked] => ifSignal(s)
       case s: EventStream[A @unchecked] => s
+      case _ => throw new Exception("All Observables must extend EventStream or Signal")
```
raquo, line 74: This should never happen but I can't make `Observable` sealed to prove it.

yurique: Yeah, I had to do this in the other branch as well. 
It's a warning, and the `sbt-tpolecat` plugin makes warnings fatal, but we don't need to have it. Although I prefer warnings like this to be caught :)

lolgab: Maybe `Exception` should instead be an exception extending `RuntimeException` like `IllegalStateException` or `RuntimeException` itself?

raquo: Eh, I don't really see the point. It's not the kind of exception that we expect people to actually run into / catch, it's just to clear the warning. I try to keep Java-native stuff out of my frontend libraries when it's not needed so I think I'll pass on that.

lolgab: In Java terms it's the `RuntimeException` the "correct" exception to use when you want to represent a programming error (not something you can recover from using a catch clause). But I see your point that you don't want java-native stuff in a frontend JS library. Makes total sense.

PR #114, opened 2023-11-16
Add `EventStream.fromPublisher`

armanbilge:
And effectively implementing (part of) the [Reactive Streams](https://www.reactive-streams.org/) spec.

In practice, this makes it possible to convert an `fs2.Stream` to an `EventStream` via the [`Flow.Publisher`](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/concurrent/Flow.Publisher.html) interface. Monix and ZIO also interop with Reactive Streams, although this is not cross-published for Scala.js yet.

Edit: [It turns out Monix also supports Reactive Streams on Scala.js!](https://github.com/monix/monix/blob/2faa2cf7425ab0b88ea57b1ea193bce16613f42a/monix-execution/js/src/main/scala/org/reactivestreams/Publisher.scala)

### DIFF
--- a/src/main/scala/com/raquo/airstream/core/EventStream.scala
+++ b/src/main/scala/com/raquo/airstream/core/EventStream.scala
@@ -14,6 +14,7 @@ import com.raquo.airstream.split.{SplittableOneStream, SplittableOptionStream, S
 import com.raquo.airstream.timing._
 import com.raquo.ew.JsArray
 
+import java.util.concurrent.Flow
 import scala.annotation.unused
 import scala.concurrent.{ExecutionContext, Future}
 import scala.scalajs.js
@@ -346,6 +347,30 @@ object EventStream {
 
   def fromJsPromise[A](promise: js.Promise[A], emitOnce: Boolean = false): EventStream[A] = {
     new JsPromiseStream[A](promise, emitOnce)
+  }
+
+  def fromFlowPublisher[A](mkPublisher: => Flow.Publisher[A], emitOnce: Boolean = false): EventStream[A] = {
+    var subscription: Flow.Subscription = null
+    fromCustomSource[A](
+      shouldStart = startIndex => if (emitOnce) startIndex == 1 else true,
+      start = (fireEvent, fireError, _, _) => {
+        
+        val publisher = mkPublisher
+        
+        val subscriber = new Flow.Subscriber[A] {
+          def onNext(a: A) = fireEvent(a)
+          def onError(t: Throwable) = fireError(t)
+          def onComplete() = ()
+          def onSubscribe(s: Flow.Subscription) = {
+            s.request(Long.MaxValue)
+            subscription = s
+          }
+        }
+
+        publisher.subscribe(subscriber)
+      },
+      stop = _ => { subscription.cancel(); subscription = null }
+    )
   }
 
   /** Easy helper for custom events. See [[CustomStreamSource]] for docs.
--- a/src/main/scala/com/raquo/airstream/core/EventStream.scala
+++ b/src/main/scala/com/raquo/airstream/core/EventStream.scala
@@ -349,7 +349,7 @@ object EventStream {
     new JsPromiseStream[A](promise, emitOnce)
   }
 
-  def fromFlowPublisher[A](publisher: Flow.Publisher[A], emitOnce: Boolean = false): EventStream[A] = {
+  def fromPublisher[A](publisher: Flow.Publisher[A], emitOnce: Boolean = false): EventStream[A] = {
     var subscription: Flow.Subscription = null
     fromCustomSource[A](
       shouldStart = startIndex => if (emitOnce) startIndex == 1 else true,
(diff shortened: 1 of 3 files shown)


src/main/scala/com/raquo/airstream/core/EventStream.scala
```diff
@@ -348,6 +349,30 @@ object EventStream {
...
+          def onComplete() = ()
+          def onSubscribe(s: Flow.Subscription) = {
+            s.request(Long.MaxValue)
+            subscription = s
+          }
+        }
+
+        publisher.subscribe(subscriber)
+      },
+      stop = _ => { subscription.cancel(); subscription = null }
```
yurique, line 372: `if (subscription != null)`? :)

armanbilge: Can that happen? Does Airstream not guarantee that every `stop` is preceded by a `start`?

armanbilge: Hmm, I guess a mis-behaved `Publisher` could refuse to call `onSubscribe`. Fair point, better be safe :)

yurique: > guarantee that every stop is preceded by a start
¯\_(ツ)_/¯  
I don't trust no guarantees 🤣

raquo:
Looks great, thank you! I should be able to include this in the next version, just need to write some docs on how to use it with FS2.

armanbilge:
> I just need to write some docs on how to use it with FS2.

I've been thinking about this. After chatting with @BalmungSan we realized that we need to add one more API on the FS2 side. Interop is possible with the current APIs but I think it would require `flatMap`ping `EventStream`s which is sub-optimal. We want to expose a new "unsafe" API that gives direct access to the `Publisher` from the `fs2.Stream` without wrapping it in `Resource` or `IO`.

raquo:
Hm, I guess I'll hold off merging until that's cleared up. I'm not sure why `flatMap` would be necessary on the airstream side, but if you'll fix that need eventually, that's moot.

armanbilge:
> I'm not sure why `flatMap` would be necessary on the airstream side

Is there a way to get from a `Future[Publisher[A]]` to an `EventStream[A]` using `fromFlowPublisher` with the current signature and not using `flatMap`?

raquo:
> Is there a way to get from a Future[Publisher[A]] to an EventStream[A] using fromFlowPublisher with the current signature and not using flatMap?

No.

To clarify, _unnecessary_ flatMap-s in Airstream are bad because they creates unnecessary transactions (which can cause FRP glitches), but this particular flatMap, although unnecessary, does not _increase_ the potential of FRP glitches, I think: since the events originate outside of Airstream (in FS2), Airstream will emit them in separate transactions anyway, even without flatMap.

Eliminating the need for a flatMap is still beneficial though: less boilerplate, and better efficiency.

armanbilge:
> Hm, I guess I'll hold off merging until that's cleared up.

FWIW I don't think it is necessary to hold this PR. The API here is solid, and interop is possible today. We can document as-is, and update the docs once FS2 adds the new API, without requiring a new Airstream release.

armanbilge:
After https://github.com/typelevel/fs2/pull/3346 is released, interop will be as simple as
```scala
import cats.effect.unsafe.implicits._ // implicit IORuntime
EventStream.fromFlowPublisher(fs2Stream.unsafeToPublisher())
```

src/main/scala/com/raquo/airstream/core/EventStream.scala
```diff
@@ -348,6 +349,31 @@ object EventStream {
     new JsPromiseStream[A](promise, emitOnce)
   }
 
+  def fromFlowPublisher[A](publisher: Flow.Publisher[A], emitOnce: Boolean = false): EventStream[A] = {
```
armanbilge, line 352: Bikeshed?

```suggestion
  def fromPublisher[A](publisher: Flow.Publisher[A], emitOnce: Boolean = false): EventStream[A] = {
```

raquo: 👍 Fine by me, I don't think I'll ever have a `Publisher` type in Airstream

armanbilge:
[FS2 v3.9.4](https://github.com/typelevel/fs2/releases/tag/v3.9.4) was released with the new interop APIs.

raquo:
@armanbilge Great, thanks, and for the docs too, I'll merge this PR for the next 17.x release

raquo:
@armanbilge Merged now, thanks again!

I wanted to rebase this on latest master, add a minor commit to this, and squash it all, but I couldn't figure out how to do that while preserving this PR and without bothering yourself for such menial work, so I've done all that in #122. The resulting commit is [here](https://github.com/raquo/Airstream/commit/fac399df615fcf02245094babfcc753f73115a51). No API changes, just moved things around a bit.

Issue #103, opened 2022-10-18
splitByType / splitByEnum


raquo:
What if you have a `Signal[MyEnum]` or `Signal[MyGADT]`, and you want to split the signal by **type**, much like the normal `split` operator splits the signal by **value**?

[Waypoint](https://github.com/raquo/waypoint) has a useful SplitRender feature that works like this:

```scala
val splitter = SplitRender[Page, HtmlElement](router.$currentPage)
  .collectSignal[UserPage] { $userPage => renderUserPage($userPage) }
  .collectStatic(LoginPage) { div("Login page") }
 
// this method requires a Signal of a specific page type, and SplitRender is able to provide it
def renderUserPage($userPage: Signal[UserPage]): Div = {
  div(
    "User page ",
    child.text <-- $userPage.map(user => user.userId)
  )
}
 
val app: Div = div(
  h1("Routing App"),
  child <-- splitter.$view // get the signal.
)
```

I could potentially move this functionality into Airstream, but it's arguably... not good enough. The biggest downside is the lack of exhaustiveness matching, that is, you can call `.$view` any time to get the signal, even if you didn't handle all the cases. I also dislike the need to call `$view` at all, perhaps I will hide it with an operator signature like def `splitByType(f: SplitRender => SplitRender): Signal[Out] = f(SplitRender(this)).$view` (various type params omitted for brevity).

@felher suggested using Scala 3 metaprogramming for this, and although my design goal is to keep Laminar & Airstream simple, I think exhaustive matching on subtypes is actually a good use for this, that is, the metaprogramming implementation will probably be the simplest most straightforward one.

I will not be using any third party macro libraries though, plain Scala only, which means Scala 3 only, for the sake of reducing dev and maintenance effort.

For reference, here is @felher's implementation of `splitByEnum`: https://gist.github.com/felher/5515eb1124268b0e10eadc78778f49a8 – it's short, and even though I have zero experience in macros, I can still understand what it does at a high level. I even dare say this might be good to include in Airstream as-is.

If we're doing this, we should probably also support a more complicated `splitByType` use case, which, similar to Waypoint's `SplitRender`, would let the user split type hierarchies on arbitrary subtypes, while providing exhaustive matching. I don't know how hard it is to implement with Scala 3 metaprogramming facilities, I need to find the time to learn this whole thing.

Any help / advice is welcome in this direction.

HollandDM:
I do have some ideals about this, and would like to help.
[...]

PR #116, opened 2024-01-27
Add `SplitByType`s observables

HollandDM:
As discussed in #103, this is the implementation of `SplitByType` macro for observable.
The macro will expand into `match/case` definition, so it can takes advantage of exhaustive and reachable checking of the compiler.


src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
```diff
@@ -0,0 +1,227 @@
...
+  )(
+    using quotes: Quotes
+  ): Expr[(Int, Any)] = {
+    import quotes.reflect.*
+
+    val caseExprList = exprOfListToListOfExpr(caseListExpr)
+
+    val allCaseDefLists = caseExprList.reverse.zipWithIndex.flatMap { case (caseExpr, idx) =>
+      caseExpr.asTerm match {
+        case Block(List(DefDef(_, _, _, Some(Match(_, caseDefList)))), _) => {
```
kitlangton, line 213:
```suggestion
        case Lambda(_, Match(_, caseDefList)) => {
```

There's actually this little helper built into the `reflectModule` that matches the (awkwardly verbose) AST for anonymous functions. 👍 Awesome macro, btw!

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
```diff
@@ -0,0 +1,220 @@
...
+   *
+   * ```scala
+   *  MatchSplitObservable.build(fooSignal, ({ case baz: Baz => baz }) :: ({ case Bar(Some(str)) => str }) :: Nil, handlerMap)
+   * ```
+   *
+   * This is important, because `({ case baz: Baz => baz }) :: ({ case Bar(Some(str)) => str }) :: Nil` saves all the infomation needed for the macro to expands into match case definition
+   */
+  final class MatchSplitObservable[Self[+_] <: Observable[_] , I, O] {
+    throw new UnsupportedOperationException("splitMatch without toSignal/toStream is illegal")
+  }
```
HollandDM, line 69: This class act as a type only. [DuckTape](https://github.com/arainko/ducktape/blob/e001d4a4edb84f64d38c698df6e93ae78cedb706/ducktape/src/main/scala/io/github/arainko/ducktape/BuilderConfig.scala) has a trick to create these kind of type (`opaque TypeName = Unit`), but I cannot seem to integrate it into this code. Will have to dig more into this.

HollandDM: Split type definition into a dedicated file seem to work.

ngbinh:
@kitlangton @raquo is there anything left to do with this PR? I think it is quite useful

raquo:
It's definitely very useful, thanks for your work on this! I am yet to properly review this, but on the high level it looks good.

Sorry for such a delay, this was just too much for me to include in v17. I plan to merge this when I start working on the next Laminar release, likely around September-October. Since all this ties into Airstream with extension methods, I assume that you can use this code privately for now.

src/test/scala/com/raquo/airstream/split/SplittableTypeSpec.scala
```diff
@@ -0,0 +1,474 @@
...
+  
+  it("split match signal") {
+    val effects = mutable.Buffer[Effect[String]]()
+
+    val myVar = Var[Foo](Bar(Some("initial")))
+
+    val owner = new TestableOwner
+
+    val signal = myVar.signal
+      .splitMatch
```
raquo, line 33: Since this test file uses the new Scala 3 only features, the test file itself should be Scala 3 only

raquo:
@ngbinh Sorry for such a long delay on this, it's definitely a very useful feature, the blame is entirely on my availability and schedule. I feel bad for letting such a good contribution go unmerged for so long. The reason is, I'm currently working overdrive on an unrelated project, and won't be able to spend a significant time on OSS until late October, maybe even early November. I'm itching very hard to publish this and other new features once I'm able to dedicate the time. Merging this PR will be one of my first priorities when I get back into it.

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
```diff
@@ -0,0 +1,185 @@
...
+    parentObservable.matchStreamOrSignal(
+      ifStream = _.splitOne(_._1) { case (idx, (_, b), dataSignal) =>
+        val bSignal = dataSignal.map(_._2)
+        handlerMap.apply(idx).apply(b -> bSignal)
+      },
+      ifSignal = _.splitOne(_._1) { case (idx, (_, b), dataSignal) =>
+        val bSignal = dataSignal.map(_._2)
+        handlerMap.apply(idx).apply(b -> bSignal)
+      }
+    ).asInstanceOf[Self[O]] // #TODO[Integrity] Same as FlatMap/AsyncStatusObservable, how to type this properly?
```
raquo, line 135: In your public API, you have `toStream` and `toSignal` methods, but you have a common implementation for them - `toSplittableOneObservable`, calling into `toSplittableOneObservable`. The last two methods require the Self type, which is where the type problem starts.

Airstream's underlying `splitOne` method is not a single method on Observable / Self, but rather we have two separate `splitOne` methods, one for streams, and one for signals. And so, they don't work with the `Self` type, requiring something like `matchStreamOrSignal` to use from a generalized Self method.

In principle, we could change Airstream to have only one unified `splitOne` method, implemented in a single SplitOneObservable value class with a Self type param, similar to e.g. BooleanObservable. However, the implementation of such a `splitOne` method would need to do something similar to `matchStreamOrSignal`, except with the output type `B <: Self`, to satisfy your method's desired output type. Perhaps add a new `matchStreamOrSignalSelf` method with such a constraint, as it may be useful in other cases as well?

There is still a minor problem with that – instead of doing `asInstanceOf`, you would need to throw an (effectively unreachable) exception like `matchStreamOrSignal` does, because the compiler thinks that we can't exhaustively match Observable as EventStream | Signal, because even though there are no other subtypes of Observable, I can not make Observable a sealed trait, because there is no way I can put the entire contents of EventStream and Signal types into one scala file for ergonomic reasons.

Still, I think the best option is to unify the `splitOne` method this way. Then you can call into a single splitOne method with the output Self type matching the input type – that will eliminate the need for `matchStreamOrSignal` in your macrro implementation. The next best option, if you don't want to mess with unifying `splitOne`, is to only implement `matchStreamOrSignalSelf` as I suggested, and just use that from your macros instead of `matchStreamOrSignal` – that should fix the types just as well. Lastly, as a last resort, you could also rewrite your macros to have separate implementations for signals and streams instead of one common implementation. But that actually sounds like more work, and would leave us with a bit of redundant code. NBD, but option  number 2 is probably superior in every way.

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
```diff
@@ -0,0 +1,185 @@
...
+    import quotes.reflect.*
+
+    matchSplitObservableExpr match {
+      case '{ MatchSplitObservable.build[Self, I, O]($_, Nil, $_)} =>
+        report.errorAndAbort("Macro expansion failed, need at least one handleCase")
+      case '{ MatchSplitObservable.build[Self, I, O]($observableExpr, $caseListExpr, $handlerMapExpr)} => 
+        '{toSplittableOneObservable(
+          $observableExpr
+            .map(i => ${ innerObservableImpl('i, caseListExpr) })
+            .asInstanceOf[BaseObservable[Self, (Int, Any)]],
```
raquo, line 152: Is this `asInstanceOf` actually needed? Which part of the type is unproven? Is it due to Airstream types, or something inherent in macros?

I would expect the output here to be `Self[(Int, Any)]` prior to `asInstanceOf`. I'm not sure how `(Int, Any)` relates to `O`, but I don't see how `asInstanceOf` helps with that either, since it does not mention`O`. It seems that you're using `asInstanceOf` to get `BaseObservable`, but all you need to return is Self, no? I could be missing something...

src/test/scala-3/com/raquo/airstream/split/SplittableTypeSpec.scala
```diff
@@ -0,0 +1,474 @@
...
+    )
+
+    effects.clear()
+
+    myVar.writer.onNext(Bar(None))
+
+    effects shouldBe mutable.Buffer(
+      Effect("result", "Res(initial)"),
+      Effect("update-child", "Bar-null")
+    )
```
raquo, line 86: I was really confused by these assertions. Intuitively, it does not make sense to expect the result to be "initial" when you have emitted "null" – at this point, "initial" is a stale value already.

I now see the root of my confusion – you're returning `Res(<initialValue>)` from the split callbacks, such as `Res(str)` or `Res(baz)`. Whereas my other tests return the `key` that the callback is for, e.g. see the callbacks returning `Bar(key)` in `SplitsignalSpec` – the resulting assertions are more intuitive there.

To help me keep my sanity, can you please replace usages of `Res(<initialvalue>)` with `Res(<key>)`, for example `Res("Bar")` instead of `Res(str)`, and `Res("Baz")` instead of `Res(baz)` above – throughout all of these new tests.

The real `key`-s used by the macros are hidden, and are integers, so you should make up semantic string keys that would basically act as debug labels to show which of the `handleCase`-s the callback belongs to, like "Bar" and "Baz" in my examples above.

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
HollandDM: `$observableExpr.map(i => ${ innerObservableImpl('i, caseListExpr) })` is indeed a `Self[(Int, Any)]`, but I'm not returning this, I'm actually feeding this into `toSplittableOneObservable` function call, and in that function I need `matchStreamOrSignal`, so `Self` is not enough.

src/test/scala-3/com/raquo/airstream/split/SplittableTypeSpec.scala
```diff
@@ -0,0 +1,474 @@
...
+        effects += Effect("init-child", s"Bar-$str")
+        // @Note keep foreach or addObserver here – this is important.
+        //  It tests that SplitSignal does not cause an infinite loop trying to evaluate its initialValue.
+        strSignal.foreach { str =>
+          effects += Effect("update-child", s"Bar-$str")
+        }(owner)
+
+        Res(str)
+      }
+      .handleCase { case baz: Baz => baz } { case (baz, bazSignal) =>
```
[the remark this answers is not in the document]
HollandDM: Yes, I think it can be done, I'll have an update for this.

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
```diff
@@ -0,0 +1,338 @@
...
+  ) {
+    inline def handleCase[A, B, O1 >: O](inline casePf: PartialFunction[A, B])(inline handleFn: ((B, Signal[B])) => O1) = ${
+      handleCaseImpl('{ matchSplitObservable }, '{ casePf }, '{ handleFn })
+    }
+
+    inline private def handlePfType[T](inline casePf: PartialFunction[Any, T]) = ${
+      handleTypeImpl[Self, I, O, T]('{ matchSplitObservable }, '{ casePf })
+    }
+
+    inline def handleType[T]: MatchTypeObservable[Self, I, O, T] = handlePfType[T] { case t: T => t }
```
HollandDM, line 74: The reason I need to take this "detour" is that, originally this should be `handleType[T][O1 >:O](inline handleFn: ((T, Signal[T])) => O1)`. `O1>:O` here is very important because it helps refining the output type from the ground "Nothing" up to the output type we want. Sadly Scala 3 right now can only do `handleType[T, O1 >:O]`, and this force user to explicitly named the output type of the `handleFn` function.
Until [SIP-47](https://docs.scala-lang.org/sips/clause-interleaving.html) is widely adopted, we will have to stick into this detour

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
HollandDM: This note was more of a todo/reminder rather than a question that needs answer right away. I'm fine with the way it is right now, but I can try to solve this if you think it is necessary.

src/test/scala-3/com/raquo/airstream/split/SplittableTypeSpec.scala
HollandDM: Fair enough, I'll changes these test

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
raquo: Eh, it's fine as-is, I'll try to get rid of this asInstanceOf if / when I unify the splitOne method.

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
raquo: Do I understand correctly that calling `handleType[Foo[A]]` is equivalent to the user writing `handleType { case f: Foo[A @unchecked] => f }` – that is, the `Foo` will be checked, but `A` will be unchecked? That's fine and that's what I would expect, given type erasure, just wanted to double check.

src/test/scala-3/com/raquo/airstream/split/SplittableTypeSpec.scala
```diff
@@ -0,0 +1,474 @@
...
+    val myVar = Var[Foo](Bar(Some("initial")))
+
+    val owner = new TestableOwner
+
+    val signal = myVar.signal
+      .splitMatch
+      .handleCase {
+        case Bar(Some(str)) => str
+        case Bar(None) => "null"
+      } { case (str, strSignal) =>
```
raquo, line 37: I just noticed that the public API requires the user to provide a function expecting a single argument – a tuple of `(A, Signal[A])` – does it have to be like this, instead of accepting two arguments, `A` and `Signal[A]`?

The single tuple argument basically requires the end user to provide a partial function using `case`, where a total function is expected. If the function accepted two arguments, we could write the same but without the second `case`:

```
      .handleCase {
        case Bar(Some(str)) => str
        case Bar(None) => "null"
      } { (str, strSignal) =>
        ...
      }
```

raquo:
If it's not too much to ask, could you please add another helper – `handleValue[A, B](value: A)(result: B) = handleCase { case a if a == value => a } { _ => result }`. This would help transition Waypoint to use these new methods.

raquo:
Sorry, correction for my `handleValue` request – it won't actually work for `object`-s that way, the match would not be exhaustive. I guess it needs to be something like:

`handleValue[A: ValueOf, B](value: A)(result: B) = handleCase { case value => value } { _ => result }`

`ValueOf` is Scala 3 type class evidencing that the value is a singleton (e.g. an `object`, or a literal type). [Here](https://github.com/raquo/Waypoint/pull/19/files#diff-72e107b3cc8eec584380b80e861f353e3c2f181372f6925272f0c5a8acfd339dR484-R488) is a non-macro usage of it, for example.

Basically, something like this:

```scala
sealed trait Foo
case class Bar(a: Int) extends Foo
object Baz extends Foo

val x: Foo = Baz

x match {
  case Bar(a) => ...
  case Baz => ...
}
```

Should translate into:

```scala
  .handleType[Bar] { (_, _) => ... }
  .handleValue(Baz)( ... )
```

src/test/scala-3/com/raquo/airstream/split/SplittableTypeSpec.scala
HollandDM: It can be changed, it's just an old code from initial implementations that slip through

HollandDM:
I'll have a look at `handleValue`, technically, it should be implemented similar to `handleType` (with the "detour" thing) so that user doesn't have to provide `B` explicitly.

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
HollandDM: Yeah I believe so, the end result of this macro will create a simple `match/cases` code, before any transformation happens. So it should follow the normal behavior of scala

src/main/scala-3/com/raquo/airstream/split/MatchValueObservable.scala
```diff
@@ -0,0 +1,52 @@
...
+ * ```scala
+ * MatchSplitObservable.build(
+ *   fooSignal,
+ *   ({ case Tar => Tar }) :: Nil,
+ *   handlerMap
+ * )
+ * ```
+ */
+
+opaque type MatchValueObservable[Self[+_] <: Observable[_], I, O, V0, V1] = Unit
```
HollandDM, line 37: `V0` and `V1` should never be different, but we cannot just do `MatchValueObservable[Self[+_] <: Observable[_], I, O, V]` because after type erasure, compiler cannot differentiates `MatchValueObservable[Self[+_] <: Observable[_], I, O, V]` from `MatchTypeObservable[Self[+_] <: Observable[_], I, O, T]`.

src/main/scala-3/com/raquo/airstream/split/SplittableTypeMacros.scala
```diff
@@ -0,0 +1,417 @@
...
+      handleTypeApplyImpl('{ matchTypeObserver }, '{ handleFn })
+    }
+  }
+
+  extension [Self[+_] <: Observable[_], I, O, V0, V1](inline matchValueObservable: MatchValueObservable[Self, I, O, V0, V1]) {
+    inline private def deglate[O1 >: O](inline handleFn: (V1, Signal[V1]) => O1) = ${
+      handleValueApplyImpl('{ matchValueObservable }, '{ handleFn })
+    }
+
+    inline def apply[O1 >: O](inline handleFn: Signal[V1] => O1): MatchSplitObservable[Self, I, O1] = deglate { (_, vSignal) => handleFn(vSignal) }
```
HollandDM, line 96: difference in signature (`Signal[V1] => O1` here) means we need another (quite redundant IMHO) dummy class and dummy "detour".

src/test/scala-3/com/raquo/airstream/split/SplittableTypeSpec.scala
```diff
@@ -0,0 +1,474 @@
...
+        effects += Effect("init-child", s"Bar-$str")
+        // @Note keep foreach or addObserver here – this is important.
+        //  It tests that SplitSignal does not cause an infinite loop trying to evaluate its initialValue.
+        strSignal.foreach { str =>
+          effects += Effect("update-child", s"Bar-$str")
+        }(owner)
+
+        Res("Bar")
+      }
+      .handleType[Baz] { (baz, bazSignal) =>
```
HollandDM, line 47: `handleType`

src/main/scala-3/com/raquo/airstream/split/MatchValueObservable.scala
raquo: TBH the implications of this are over my head, but it sounds like it's just an unfortunate implementation detail that we have to deal with. In that case, fine by me, _I_ certainly don't have better ideas on this.

I think we're all good now, thanks for all the updates!

I am super pumped for these new operators, this will be one of the first things that I will merge for 18.x, proooobably in a few weeks (but apologies in advance if my 18.x gets delayed a bit, honestly, it's possible, but I'll do my best to carve out some time for it asap).

HollandDM: This ambiguity is due to the usage of opaque types. I changed them back to value classes to remove this implementation hack. It should be easier to read and understand now.

HollandDM:
@raquo should we have another variant for `Seq[Bar]`.
In my code, I use to create something like `splitEithers`, which is kind of like a `split` & `splitEither` hybrid, it split a `Seq[Either[A, B]]` in to `Seq[render(Signal[A]) | render(Signal[B])]`, and use it quite alot.

raquo:
> should we also support Var[Bar] too?

@HollandDM Hmmm yeah, if it's actually possible, it would be nice...

```scala
trait Bar
case class Foo(int: Int) extends Bar
case obect Baz extends Bar
```

To mirror the normal Var.split API, ideally we would want to provide an API like the following, I guess?

```scala
val barVar: Var[Bar] = ???
barBar
  .matchSplit
  .handleCase { case Foo(num) => num } { (num: Int, numVar: Var[Int]) => ... }
  .handleValue(Baz) { (bazVar: Var[Baz]) => ... }
```

So then, we would need to e.g. create such a Var[Int] that would write Foo(int) into barVar – but I don't think that possible to automatically create such a Var, just from `case Foo(num) => num` – which is an arbitrary function.

So then, it looks like we would need to add another parameter, e.g.:

```scala
barBar
  .matchSplit
  .handleCase { case Foo(num) => num } { num => Foo(num) } { (num: Int, numVar: Var[Int]) => ... }
  .handleValue(Baz) { (bazVar: Var[Baz]) => ... }
```

Which is a bit verbose, but ultimately fine – it would be similar to how we have two params in the `zoom` method.

raquo:
PR for such a `splitEithers` method would be welcome. I guess if the user provides e.g. `_.id` as the splitting key, you would need to use `v => (v.id, v.isRight)` as the real key under the hood, and then call either the left or the right callback depending on isRight.

I do plan to resolve #157 eventually, although I'm not sure how it relates to `splitEithers`.

HollandDM:
For `Seq.splitMatch` and `Var.splitMatchOne` , I believe we can support them. I spent some time last weekend to play around the macros and quite confident in the results. The only remain things are the signatures of these methods.

In `Seq.splitMatch` (I think I'll change the current `splitMatch` into `splitMatchOne`, as its closer to it underlying mechanism), I think we could go with:
- `handleCase { case A => B } { B => Key } { (B, Signal[B]) => O }` or `handleCase { case A => B }(key: B => Key, project: (B, Signal[B]) => O)`.
- `handleType[T](T => Key)((T, Signal[T]) => O)` (looked pretty similar with the current `split` signature) or `handleType[T](key: T => Key, project: (T, Signal[T]) => O)`.
- `handleValue(Bar)(=> Key)(Signal[Bar] => O)` or `handleValue(Bar)(key: => Key, project: Signal[T] => O)`.

For `Var.splitMatch`, in the codebase right now we only support `split`, not `splitOne`, so the signature would be more verbose if we introduce them. In case you decided to support `splitOne` for `Var`, I think `(A => B)(B => A)` is unavoidable.

Also, I think we will need a way to let users pass their `distinctCompose` and `DuplicateKeysConfig` into these methods. What do you think?

> I do plan to resolve #157 eventually, although I'm not sure how it relates to splitEithers.
This is more than Laminar only solution, as you can do:
```scala
children <-- eitherSeq.split(eitherToKey) { (_, _, eitherSignal) =>
   child <-- eitherSignal.splitEither(
       left = ???,
       right = ???
   )
}
```

raquo:
@HollandDM So uh this is a lot of stuff at once... sorry if I mixed anything up.

This new `splitMatch` that you're proposing (let's call it `splitMatchSeq` to differentiate from the current `splitMatch` aka `splitMatchOne` in discussions) would basically be a more generalized version of `splitEithers`, it would be to `splitMatch` what `splitEithers` would be to `splitEither`, and `split` to `splitOne`, ok. I like the idea.

Let's ignore the `Var` splitting component to this for now – one thing at a time.

How to define split keys in `splitMatchSeq`? One option is to define it as a single "global" config when calling `splitMatchSeq`:

 ```
 signalOfFoos: Signal[List[Foo]] = ???
 
 signalOfBars
   .splitMatchSeq(
     key = {
       case Foo(num) => num
       case Foo2(idStr) => idStr
       case Bar => ()
     },
     distinctCompose = ???,
     duplicateKeys = ???
   )
   .handleCase { case Foo(num) => num } { (num: Int, numSignal: Signal[Int]) => ... }
   .handleType[Foo2] { (foo2: Foo2, foo2Signal: Signal[Foo2])  => ... }
   .handleValue(Bar) { (barSignal: Signal[Bar.type]) => ... ]
 ```
 
This introduces some redundancy between the definition of `key` and the `handleCase`-s, however this simplifies the API of `handleCase` and other helpers, and lets us expand this syntax for Var-s as well (where we'll need to provide another `B => A` zoomOut callback). I think adding key callbacks to each `handleCase` would simply be too unwieldy – too many callbacks in one method invocation.
 
Perhaps the shared key definition could be simplified if the common Foo type had a property like `id` that all subtypes shared.
 
The real key used by `split` under the hood would need to be a tuple of `(interrnalMacroCaseNumber, userKey)`. Thus, `userKey` could be Any, the keys for different cases don't need to be of the same type, they are simply compared with `==` internally.

And so, another downside of defining the keys separately like this is that we don't get the `key` in each render callback. But we do get the initial value (`num` or `foo2` in my example), and the key can usually be trivially derived from that (often it's just `initialValue.id` or similar).

OTOH I think all of the above can be implemented just in terms of `split` similarly to how you implemented `splitMatch` in terms of `splitOne`. But if you wanted to e.g. provide a separate`distinctCompose` to each `handleCase`, I don't know if that would be possible, it may require changes to the source of SplitSignal, which would be unpleasant.

Overall I think I still to prefer this syntax over specifying the key in every case. What do you think, would it fit your use cases? Am I missing some more ergonomic concerns?

---

Regarding the naming, I think `splitMatchSeq` is a better name for this, even if it does not align with the name of the `split` operator. Me naming that method `split` was a mistake, in hindsight it should have been `splitSeq`, then we could have named `splitOne` as simply `split`, and the names of all the other helpers like `splitOption`, `splitMatch`, etc. would also be consistent. I think eventually I'll rename `split` to `splitSeq`, but for now, I don't want to break the code in the videos that mention the `split` operator, so it's ok to name these new operators inconsistently with `split` / `splitOne` (but consistently with other `split*` helpers).

HollandDM:
using a single "global" function make can make the `key`s out of sync with the `handle`s branch. E.g:
```scala
signalOfFoos
  .splitMatchSeq(
    key = {
      case Foo(num) if num > 0 => num
      case Foo2(idStr) => idStr
      case _ => ()
    },
    distinctCompose = ???,
    duplicateKeys = ???
  )
  .handleCase { case Foo(num) => num } { (num: Int, numSignal: Signal[Int]) => ... }
  .handleType[Foo2] { (foo2: Foo2, foo2Signal: Signal[Foo2])  => ... }
  .handleValue(Bar) { (barSignal: Signal[Bar.type]) => ... ]
```
This is a valid syntax, but `key`s and `handleCase`s are out of sync.

Perhaps we should just separate the key definition here from the `handle*` calls, then internal macros can use the key `(interrnalMacroCaseNumber, userKey)` like you said, but with a known type for `userKey`, and we can also put the key in the `handle` callback as well.

For me personally, beside my custom `splitEithers`, I haven't find any other need for this `splitMatchSeq` yet, so my use cases is quite limited. I guess we kinda have to decide the signatures first and then see if it works well enough.

raquo:
> This is a valid syntax, but keys and handleCases are out of sync.

I see that there is room for this, but in practice the keys will depend on the type only, so there should be no value-based conditions like `if num > 0` even if they're present in the handleCase-s. And of course, using `case _ =>` when the only remaining option is known is IMO a bad practice that bypasses exhaustiveness checks, that one is really on the user. I think Scala's built-in exhaustiveness check here provides enough of a safeguard to eliminate the vast majority of bugs.

I can imagine very complex cases where you would need to pay attention to get the keys right, but realistically, those would be very rare, and I'm ok to take the tradeoff of putting an ergonomic burden on those rare use cases to get a simpler API for the 99% of typical uses, because we don't have a perfect alternative that offers clean syntax and good ergonomics for the typical use cases, especially when we consider the eventual support for splitting Var-s with this operator.

> but with a known type for userKey, and we can also put the key in the handle callback as well.

A shared type for userKey would only be useful in a narrow set of cases, but e.g. for an Either, it could easily end up Int | String, or more realistically it would resolve to Any without explicit type ascriptions. Ideally we would want key types to be specific to each handleCase, but that is not possible without moving key callbacks into individual handleCase calls.

I think in practice we can live without providing the key to the handle callback. Usually the key would be something obvious like `_.id`, so it's not a big deal just to get it from the initial value provided by the callback.

HollandDM:
> A shared type for userKey would only be useful in a narrow set of cases, but e.g. for an Either, it could easily end up Int | String, or more realistically it would resolve to Any without explicit type ascriptions. Ideally we would want key types to be specific to each handleCase, but that is not possible without moving key callbacks into individual handleCase calls.

I think Its quite impossible for key to be specific for each `handle*` method if we use this syntax, that why I suggest we don't associate key with `handle*` at all. For a `M[I]` input type, we make `key` to be `I => K`, and proceed with the `handle*` methods like this:

```scala
signalOfFoos
  .splitMatchSeq(
    key = Foo => K,
    distinctCompose = ???,
    duplicateKeys = ???
  )
  .handleCase { case Foo(num) => num } { (num: Int, numSignal: Signal[Int]) => ... }
  .handleType[Foo2] { (foo2: Foo2, foo2Signal: Signal[Foo2])  => ... }
  .handleValue(Bar) { (barSignal: Signal[Bar.type]) => ... ]
```

I'm not too concern with initial key being provided or not. I just want to know if `key`s need to follow `handle*`s definitions or not.

raquo:
Ok, yes, I think we're on the same page.

If I understand correctly – no, the user-provided keys don't need to follow handleCase-s. Any `Foo => K` function will work, e.g. if Foo has a common id field, the key function could just be `_.id`. The only constraint on the keys is as always – each item in the Seq must have a unique key, but here, this constraint is even relaxed a bit – each item in the Seq must have a key that is unique **only within the subset of items in the Seq that are rendered with same handleCase** – because the real key used under the hood will have a macro-generated `handleCaseIndex` in it: `(handleCaseIndex, userProvidedKey)`. Is that what you were asking?

Also, just wanted to mention that since the type K in `Foo => K` won't be exposed anywhere, it could just be Any, no need for a type param to encode it.

HollandDM:
Yeah, I just want to clarify the key point before committing, I see that we agree the `key` could be any function of type `Foo => K`.
I'll update the ticket to include `splitMatchSeq` implementations and testings first (ideally I would follow SplitSpec).
About `Var`, I think we can continue after `splitMatchSeq` is good to merge.

raquo:
Sounds good, thank you for all this 🙏

HollandDM:
@raquo `splitMatchSeq` is pushed, please have a review.
I also changed the old `splitMatch` to `splitMatchOne` for user to distinguish them better.
The spec of `splitMatchSeq` follow `SplitSignalSpec` style, but I'm not copy/paste all of them because of redundancy, so please let me know if you want to include a specific test.

ngbinh:
@raquo is there anything else left to do for this PR? Thanks

raquo (approved):
All is good, thanks! Super stoked to have this feature. I'm finally able to slowly start work on the next version, so I'll be merging this soon.

raquo:
@HollandDM Sorry I was too eager to merge this, but now I see there are some issues yet remaining.

_Note – I did not expect to need more changes, so I've squash-merged the PR, as I usually do to clean up git history, but because of that you'll need to force rebase on Airstream's master to reconnect to that history. Sorry for the inconvenience :|_

#### Using too many cases at once slows down the compiler, eventually resulting in an OOO crash

To reproduce, add this test to SplitMatchOneSpec:

```scala
it("xxx") {
    val myVar = Var[Foo](Bar(Some("initial")))
    val signal = myVar.signal
      .splitMatchOne
      .handleCase { case Bar(Some(str)) => str } { (str, strSignal) => () }
      .handleCase { case Bar(Some(str)) => str } { (str, strSignal) => () }
      // !!! TODO: To crash the compiler, insert the same handleCase line 20-40 times here
      .handleCase { case Bar(Some(str)) => str } { (str, strSignal) => () }
      .handleCase { case Bar(Some(str)) => str } { (str, strSignal) => () }
      .toSignal
  }
```

For me it usually takes 10-20 seconds for the compiler to crash. Depends on CPU and heap I guess. If it doesn't crash for you, you may need to add more lines. It will still be slow though, and it may give different warnings:

`In the last 10 seconds, 5.333 (53.8%) were spent in GC. [Heap: 0.00GB free of 1.00GB, max 1.00GB] Consider increasing the JVM heap using `-Xmx` or try a different collector, e.g. `-XX:+UseG1GC`, for better performance.`

`scala.tools.asm.MethodTooLargeException: Method too large: com/raquo/app/views$package$.<clinit> ()V`

I ran into this issue when trying to convert [these waypoint cases](https://github.com/raquo/laminar-full-stack-demo/blob/bf97b0fd5269f6b54f51cf211ae5819315062591/client/src/main/scala/com/raquo/app/views.scala#L13-L31) in laminar-demo app.


#### The new Airstream tests compile with Scala 3.3.3, but not with 3.3.4 or higher.

- I get lots of "Macro expansion failed, please use `handleCase` instead" errors.

- Also, Scala 3.3.4 starts throwing these warnings:

```
[warn] -- [E197] Potential Issue Warning: /Users/raquo/code/scala/airstream/src/main/scala-3/com/raquo/airstream/split/SplitMatchSeqMacros.scala:49:148
[warn] 49 |    inline def handleValue[V](inline v: V)(using inline valueOf: ValueOf[V]): SplitMatchSeqValueObservable[Self, I, K, O, CC, V] = handlePfValue[V] { case _: V => v }
[warn]    |                                                                                                                                                    ^
[warn]    |   New anonymous class definition will be duplicated at each inline site
[warn]    |----------------------------------------------------------------------------
[warn]    | Explanation (enabled by `-explain`)
[warn]    |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[warn]    | Anonymous class will be defined at each use site, which may lead to a larger number of classfiles.
[warn]    |
[warn]    | To inline class definitions, you may provide an explicit class name to avoid this warning.
[warn]     ----------------------------------------------------------------------------
[warn] four warnings found
```

  - The warnings were introduced here: https://github.com/scala/scala3/issues/16723
  - The Scala issue contemplates fixes [like this](https://github.com/tarao/record4s/pull/114/files), but if I understand correctly, I think the compiler is complaining that `{ case _: V => v }` will create a new anonymous class every time the user calls `handleValue(value)`. So I'm kind of not sure if this is a real problem in our case. Is there a way to see the scala code generated by macros, to confirm that it's generating the code we expect?

PR #118, opened 2024-02-09
Update README.md

rgwilton:
Change PeriodicEventStream to PeriodicStream

Definition of EventStream.periodic is:

```
  def periodic(
    intervalMs: Int,
    resetOnStop: Boolean = false
  ): PeriodicStream[Int] = {
    new PeriodicStream[Int](
      initial = 0,
      next = eventNumber => Some((eventNumber + 1, intervalMs)),
      resetOnStop = resetOnStop
    )
  }
```

hence, I assume that the underlying class name has changed.


raquo:
Thanks! Indeed, I removed "Event" from event stream class names a while ago.

PR #127, opened 2024-08-05
add scalafmt

doofin:
add scalafmt and just formatted all source files


raquo:
Adding scalafmt would be a good idea in general, but whoever does it needs to also write the scalafmt configuration style to match the current prevailing code formatting style in the codebase. I don't expect that scalafmt can do a 100% match, but it should come pretty close without relying on any special `formatter:off` comments.

doofin:
thanks, do you find somewhere scalafmt didn't format nicely so I can take a look? Seems the default style looks ok

raquo:
Yes, this PR has +10K - 5K line of code changes, almost all of them are undesirable to me. I can tolerate a small number of changes in formatting here and there for the sake of easier collaboration with scalafmt, but I don't want to reformat the whole codebase in some other style, just because it happens to be the default scalafmt preset. Aside from my formatting preferences, massive changes like this also erase git history, which is important to me when dealing with complex issues.

doofin:
understand, I will close this and just format a few number of files while adding scalafmt

PR #129, opened 2024-08-21
New: `.splitByIndex` and `.splitOption` for vars

phfroidmont:
Here is my attempt at adding some common operators for Var. I mostly copied and adapted what was already done for Signal.
The tests are very much a copy of the Signal equivalent. I'd be happy to add more if you want to cover more specific cases.

### DIFF
--- a/src/main/scala/com/raquo/airstream/extensions/OptionVar.scala
+++ b/src/main/scala/com/raquo/airstream/extensions/OptionVar.scala
@@ -1,0 +1,59 @@
+package com.raquo.airstream.extensions
+
+import com.raquo.airstream.core.Signal
+import com.raquo.airstream.state.{LazyDerivedVar, LazyStrictSignal, Var}
+import com.raquo.airstream.split.DuplicateKeysConfig
+
+class OptionVar[A](val v: Var[Option[A]]) extends AnyVal {
+
+  /** This `.split`-s a Var of an Option by the Option's `isDefined` property.
+    * If you want a different key, use the .split operator directly.
+    *
+    * @param project - (initialInput, varOfInput) => output
+    *                  `project` is called whenever the parent var switches from `None` to `Some()`.
+    *                  `varOfInput` starts with `initialInput` value, and updates when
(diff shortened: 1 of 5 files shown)


raquo:
Thanks, looks good! 👍 

The existing Var split tests in Airstreamn only covered the reading (signal) part, but for var splitting we should also cover the writing part, i.e. verify that **updating** those derived vars actually works as expected.

This is rather annoying to do in Airstream without access to Laminar's lifecycle management, so for regular Var splitting I implemented those writing tests in the Laminar repo. If you could look at the **"standard child Var behaviour"** test in **Laminar's** SplitVarSpec.scala, and add a similar one for `splitByIndex` (yes, to the same SplitVarSpec.scsla test file in Laminar repo), that would be great.

Don't worry about other tests in that file, no need to re-test those things for `splitByIndex`.

Basically what that "standard child Var behaviour" test is doing, is splitting the var, and then inside of it, creates child elements that record a bunch of events into the log when they're created/updated/etc., then we verify that the expected events were logged, then we simulate user clicks on those child elements (`clickOnItem`), **which triggers Var updates** (via the split mechanism that we want to test, so that's the important part), then we again test that the DOM looks as we expect, and that the expected events happened to achieve that.

So basically you would take that one test, duplicate it, replace `split` with `splitByIndex` and I guess it's mostly the same after that, you'll just be using the index as the new id (make sure to rename references to id as index where it makes sense). 

OTOH I think we can skip also rewriting that test for `splitOption`: compared to `splitByIndex`, its implementation is much more obvious, AND the test would need more changes, so less ROI on that – but if you feel like it, go for it.

raquo:
PS Running the test in the Laminar repo would require publishLocal-ing your new version of Airstream – so your tests will pass locally, but will fail in CI, but that's fine. Please do not include the Versions.scala file with your local version, I'll just remember that it's needed.

raquo (approved):
Thanks again for this, and for the patience! I'll include it in the next version.

PR #135, opened 2024-12-04
fix: Remove the need of explicit import for `SplitByType` macros

HollandDM:
Sorry for taking so long to fix this, I was busy these days.
Macro inlining functions (which has body of ${ ... } ) are required to be in the same scope (maybe same file is good enough, but I always put them in the same object). So I just create some `deglate*` functions and call them inline.

### DIFF
--- a/src/test/scala-3/com/raquo/airstream/split/SplitMatchOneSpec.scala
+++ b/src/test/scala-3/com/raquo/airstream/split/SplitMatchOneSpec.scala
@@ -4,7 +4,6 @@ import com.raquo.airstream.UnitSpec
 import com.raquo.airstream.eventbus.EventBus
 import com.raquo.airstream.fixtures.{Effect, TestableOwner}
 import com.raquo.airstream.state.Var
-import com.raquo.airstream.split.SplitMatchOneMacros.*
 
 import scala.collection.{immutable, mutable}
 import scala.scalajs.js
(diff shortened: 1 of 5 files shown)


src/test/scala-3/com/raquo/airstream/split/SplitMatchOneSpec.scala
```diff
@@ -4,7 +4,6 @@ import com.raquo.airstream.UnitSpec
 import com.raquo.airstream.eventbus.EventBus
 import com.raquo.airstream.fixtures.{Effect, TestableOwner}
 import com.raquo.airstream.state.Var
-import com.raquo.airstream.split.SplitMatchOneMacros.*
```
HollandDM, line 7: No more importing

src/test/scala-3/com/raquo/airstream/split/SplitMatchSeqSpec.scala
```diff
@@ -6,7 +6,6 @@ import com.raquo.airstream.eventbus.EventBus
 import com.raquo.airstream.fixtures.{Effect, TestableOwner}
 import com.raquo.airstream.ownership.{DynamicOwner, DynamicSubscription, ManualOwner, Subscription}
 import com.raquo.airstream.split.DuplicateKeysConfig
-import com.raquo.airstream.split.SplitMatchSeqMacros.*
```
HollandDM, line 9: No more importing

raquo:
Awesome, thank you very much for all your work on this! I will release 17.2.0 with this feature soon.

PR #142, opened 2025-05-21
Add mapSomeTo method to BooleanSignal and corresponding tests

cheleb:
Provide an extension that allows to rewrite:
```scala
val debugVar = Var(false)

div(
    child <-- debugVar.signal.map:
       case true =>
         div("My content")
       case false => div()
)
```

As:

```scala
div(
   styleAttr := "float: both;",
   child.maybe <-- debugVar.signal.mapToSome:
      div("my content")
```


cheleb:
Ho, too much honor, Nikita, thank you so much for all I learn every time I dig in laminar.

PR #150, opened 2026-02-27
Fix: Add binary compatibility forwarder for EventStream.recover

nguyenyou:
## Summary

- Add a single forwarding override for `recover` on `EventStream`
- Restores the Scala.js IR method descriptor that libraries compiled against Airstream 17.x expect

Fixes #149

## Context

As diagnosed by @sjrd in [this comment](https://github.com/raquo/Airstream/issues/149#issuecomment-3973941873), `RecoverOps[+Self[+_], +A]` has `Self` without the `<: Observable[_]` upper bound that `BaseObservable` has. When `recover` moved to `RecoverOps` in 18.0.0-M3, the Scala.js IR method descriptor changed (erased return type went from `Observable` to `Object`), breaking binary compatibility for downstream libraries compiled against 17.x.

The fix is minimal: only `EventStream.recover` needs a forwarder, since caliban (the affected library) only calls `.map` and `.recover` on `EventStream`, and `.map` is already overridden.

## Reproduction

https://github.com/nguyenyou/airstream-recover-bincompat-repro

## Test plan

- [x] All 309 Airstream tests pass
- [x] Verified with [minimal repro](https://github.com/nguyenyou/airstream-recover-bincompat-repro): library compiled against Airstream 17.2.1 now links successfully with the patched Airstream 18.x

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/main/scala/com/raquo/airstream/core/EventStream.scala
+++ b/src/main/scala/com/raquo/airstream/core/EventStream.scala
@@ -39,6 +39,13 @@ with DynamicImportStreamOps[A] // dynamicImport (Scala 3 only)
     new MapStream(this, project, recover = None)
   }
 
+  // BinCompat forwarder: In 17.x, recover was defined directly on EventStream.
+  // In 18.0.0-M3 it moved to RecoverOps[Self, A] where Self has no
+  // `<: Observable[_]` bound, changing the Scala.js IR method descriptor.
+  // This forwarder restores the old descriptor so that libraries compiled
+  // against 17.x (e.g. caliban) can link against 18.x.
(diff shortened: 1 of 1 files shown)


raquo:
Sorry, I don't think I'll be merging this after all. You've already worked around this locally, and nobody else has expressed interest in this. If I merge this, then yes, we'll get a bit of binary compatibility in v18 release, but then we'll have to break it again in the next release once we remove this forwarder (as I understand it). So it would just be delaying the breakage. Might as well be done with that now if it's not especially hurting anyone 🤷‍♂️

PR #151, opened 2026-03-09
Refactor: Moved and documented mapping-related operators.

SgtSwagrid:
# Refactored mapping-related operators.

1. Renamed `CoreOps` -> `MapOps` because this is what they are and this is more informative.
2. Moved `MapOps` from package `core` to `map`.
3. Moved `MapSignal` and `MapStream` from package `misc` to `map`.
4. Added ~extensive~ additional documentation to `MapOps`.


src/main/scala/com/raquo/airstream/map/MapOps.scala
```diff
@@ -0,0 +1,125 @@
+package com.raquo.airstream.map
+
+import com.raquo.airstream.core.{EventStream, Signal}
+import org.scalajs.dom.MouseEvent
```
raquo: By convention, we never import anything from `org.scalajs.dom` in Airstream / Laminar etc. We always import the `dom` package itself, so you would refer to MouseEvent as `dom.MouseEvent` in the code.

Same for the `org.scalajs.js` package – we use the `js.` prefix.

Reason is, those packages have many names in them that mirror the types in Scala / Laminar / Airstream etc. (e.g. Array / Element / Event) so we want to be very clear when the types / values we're working with are from JS land.

src/main/scala/com/raquo/airstream/map/MapOps.scala
```diff
@@ -0,0 +1,125 @@
...
+ * A base type for observables that have mapping operators such as `map` and `mapTo`.
+ *
+ * @tparam Self The kind of observable that this is,
+ *              and that will be subsequently returned by the operators herein
+ *              (e.g. [[Signal]] or [[EventStream]]).
+ *
+ * @tparam A The type of value emitted by this observable (e.g. [[MouseEvent]] or [[Int]]).
+ */
+trait MapOps[+Self[+_], +A] {
+
```
raquo, line 11: My general comment on these scaladocs is that they are way too long.
- It's too much stuff to read as a user seeking documentation
  - The most relevant & important bits of info end up being buried among trivial or only tangentially related background info
- It's too much stuff to scroll through a a user reading the source code of this file
- It's too much redundant stuff to keep updated as documentation
- You should assume that this code will be read and updated by humans, not LLMs. Repetition and verbosity has a very real cost.

For reference, compare to the original comments in CoreOps. They might be a bit light, but still MUCH closer to the desired amount of comments than the amount of comments in the new MapOps.

Some ways to make comments more compact:
- Assume basic Scala competency
  - `@param project The transformation by which emitted values are computed, based on the original value.`
    - If the user is writing Scala, they don't need the meaning of `project` callbak in a `map` method explained to them. And if they do need it, then this line alone is not sufficient explanation. So it should be omitted entirely.
  - `@tparam B The type of value emitted by the resulting observable.`
    - is not needed, the `@tparam` line can be omitted altogether
- Recognize that the user can go-to-definition for a given method and see the 20-40 characters in which it's usually implemented. We don't always need to translate these 20-40 characters of code into 200+ characters of prose text.
- Avoid providing background information that is not directly about the exact operator the scaladoc is for
  - e.g. `mapTo` scaladoc should not contain any info on how `map` behaves. At most, if you want to refer to `map`, it should say "Works like `map(_ => value)`; re-evaluates `value` on every event". You could add "see [[map]]" for easy linking – users can take the extra step.
- If you need to write something very standard in many operators – e.g. explain the handling of exceptions, please come up with a short one-liner. We can't re-explain how standard exception handling works in Airstream in every operator. My standard phrase for this is `Note: guarded against exceptions` – better versions are welcome, but must be literally a single line.
- Avoid explaining other operators that you link with `@see` – their usefulness / distinction should be obvious by name alone. If it's not, user can click on the referenced link. You also don't need to link to every single variation, just the more canonical ones. In total, something like `@see [[mapTo]], [[tapEach]]` would be sufficient for the `map` operator.
- Skip the `@return` comments for operators. They're generally self-evident given the rest of the description.

The above are good guidelines for new operators' scaladocs – please apply them there as well.

src/main/scala/com/raquo/airstream/map/MapOps.scala
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Map all emissions of this observable through a pure function `project`
```
raquo: Please don't use the term "emissions", we don't use that term in the code or in the docs. In this context the appropriate Airstream terms are "events" or "updates".

raquo (commented):
Thanks! I agree with this move & rename, but the amount of new scaladoc is out of hand. Please see code comments for details.

src/main/scala/com/raquo/airstream/map/MapOps.scala
SgtSwagrid: Fair enough. I originally figured 'emission' made some degree of sense seeing as it's the noun from "emit" which you do use, and I wanted something that wasn't Signal- or EventStream-specific. I'll switch it up.

src/main/scala/com/raquo/airstream/map/MapOps.scala
SgtSwagrid: That makes a ton of sense. In this case, I'll just remove the import as it's just for an example.

SgtSwagrid:
I guess I have a slightly different perspective vis-a-vis doc strings. I'm someone who (a) appreciates having all the information I could possibly want pop up right in my IDE when I hover over something, and (b) who has a rather large screen so is mostly undisturbed by (or unaware of the disturbance caused by) the space taken up in the file.

I've shortened all the explanations significantly, and have removed some details. Everything is now extra succinct and crisp.

raquo:
Thanks, I merged this, then polished the comments some more in https://github.com/raquo/Airstream/commit/a65086c651abfa0fde0d637b6ba7a14eed3f60b9:

- Replaced "Exceptions thrown during evaluation are caught by Airstream (see `recover()`)." with "Exceptions here are emitted as errors." – it's shorter, and is more clear about what's happening.

- Removed "Exceptions thrown during evaluation go uncaught in the current call stack." because it explains how Scala works when passing a standard (by-value) argument to a function – that's not about how _Airstream_ works. We could add this line to every method argument that isn't by-name (`: =>`), even the callbacks (an exception could happen when evaluating `project`, not just when evaluating `project(ev)`), but that wouldn't be helpful. Users are assumed to have basic Scala knowledge. I just said that it's evaluated by-value / strictly, that should be enough.

- Removed redundancies like "The resulting observable emits once each time this parent observable emits." and "to be called exactly once per event or update." (combined them together)

- Removed operators from `@see` that are not directly relevant to the current operator. If someone wants to see all the operators defined in MapOps, they can just go-to-definition.

In another commit I've also added a scalafmt rule to standardize the scaladoc asterisk spacing on what happens to be the prevailing Airstream style.

PR #152, opened 2026-03-09
Refactor: Extracted "Signal.updates()" and related methods into a separate trait.

SgtSwagrid:
# Extracted "Signal.updates()" and related methods into a separate trait.

1. Extracted `updates`, `composeUpdates`, `composeAll`, `updates()`, `composeChanges`, `changes`, and `changes()` from `Signal` into `conversions.UpdatesSignalOps`.
2. Moved `SignalFromStream` and `StreamFromSignal` from package `misc` to `conversions`.

### DIFF
--- a/src/main/scala/com/raquo/airstream/common/SingleParentSignal.scala
+++ b/src/main/scala/com/raquo/airstream/common/SingleParentSignal.scala
@@ -1,8 +1,8 @@
 package com.raquo.airstream.common
 
+import com.raquo.airstream.conversions.SignalFromStream
 import com.raquo.airstream.core.{Observable, Protected, Signal, Transaction, WritableSignal}
 import com.raquo.airstream.distinct.DistinctSignal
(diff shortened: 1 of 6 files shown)


raquo:
I'm not sure about this. Maybe it should be done, maybe not.

The *Ops traits are not free, they cost indirection, hiding methods from EventStream and Signal traits, where they should ideally exist. In case of other Ops traits, such as MapOps or DistinctOps, we get benefits from that, namely that we can reuse the same MapOps trait in EventStream, Signal, and StrictSignal, without duplicating all the `map*` operators in each of those observable types. But there is no such benefit with these .updates methods, as they are only defined for signals.

I say let's shelve this change until much later. Maybe once the dust settles on all the other operators, it will be weird to _not_ have .updates opearators in their own trait – maybe that would seem inconsistent. But I'm not there yet mentally.

PR #153, opened 2026-03-09
Feature: Added `reduceXXX()` operators.

SgtSwagrid:
The goal here is to strengthen Airstream's `scanLeft` operator (and friends). The semantics should be crystal clear, robust, unsurprising, and well-documented; in general and especially in relation to error handling and stream restarts. Common variants that the user will likely expect should be made available, and the behaviour and interface should be consistent across all of them. The operators should also be flexible enough to handle different use-cases.

This addition was largely motivated by https://github.com/raquo/Airstream/pull/147, where a desire to include new operators built on top of `scanLeft` showed that the existing semantics could be a bit inflexible and unclear. That change has now been deferred until this one can be finalised.

## *Change 1:* Refactored `Signal` and `EventSteam` so as to move existing reduction operators into traits.

Created new traits `ScanLeftStreamOps` and `ScanLeftOps` alongside the already-existing `ScanLeftSignalOps`. `ScanLeftOps` is the new common supertype for the two other types. Some repeated logic could be avoided this way.

Moved the following existing functions into the new base traits:
- `scanLeft`
- `scanLeftRecover`
- `foldLeft` (deprecated)
- `foldLeftRecover` (deprecated)

All implementations (of the above and the below) are ultimately defined in terms of `scanLeftRecover` for streams and `scanLeftGeneratedRecover` for signals, which remain the only operators here whose implementation lies directly in `Signal` and `EventStream`.

Additionally, the existing `scanLeft` and `scanLeftRecover` on `Signal` were renamed to `scanLeftGenerated` and `scanLeftGeneratedRecover` respectively (names subject to change).

## *Change 2:* Added new `reduceLeftXXX()` operators for `Signal` and `EventStream`.

See [here](https://github.com/raquo/Airstream/pull/147#issuecomment-4022406145) for a brief exposition on nomenclature for the new `reduce` operators.

See [here](https://github.com/raquo/Airstream/pull/147#issuecomment-3963227736) for a discussion of on what types these operators should be defined and what types they should return.

| Operator | Defined for | Returns | Description |
| - | - | - | - |
| `reduceLeft` | both | same as parent* | similar to `scanLeft` but without a seed value |
| ~`reduceLeftOption`~ | `EventStream` only** | `Signal` | has value `None` before first event |
| ~`reduceLeftDefault`~ | `EventStream` only** | `Signal` | has default value before first event |
| `reduceLeftRecover` | `Signal` only*** | `Signal` | uses `Try` to recover from errors |
| `scanLeft` | both | `Signal` | new variant on `Signal` has seed value instead of generator |
| `scanLeftRecover` | both | `Signal` | new variant on `Signal` has seed value instead of generator |

`*` Streams can't reduce to signals because the value is undefined before the first event.
`**` Not meaningful for signals as there is never a missing value.
`***` Technically possible for streams but implementation proved to be annoying.

## ~*Change 3:* Added new boolean flags `resetOnStop` and `skipErrors` to the above functions.~

See [here](https://github.com/raquo/Airstream/pull/147#issuecomment-3963030932) for the discussion which motivated the inclusion of these flags.

These flags are available in all of the above functions, with the exception of `XXXRecover()` where `ignoreErrors` doesn't make sense because error handling is done by the caller in these cases.

- `resetOnStop`: Whether to reset the accumulator when the observable is restarted due to all subscribers leaving. Defaults to `false` for consistency with existing behaviour.
- `skipErrors`: Whether to ignore errors, simply omitting them from the accumulation. Defaults to `false` for consistency with existing behaviour, where a single error will render the stream inoperable.

## ~*Change 4*: Added library of basic "recovery strategies" describing ways to handle errors in reductions.~

There are three kinds of error that a recovery strategy must consider:
  - Upstream errors: Those originating from the parent or another ancestor.
  - Seed errors: Those originating from computation of the initial seed value.
  - Combine errors: Those originating from the combine function.

Added three example strategies to `scan/Recover.scala` to demonstrate:
  - `Recover.keepErrors`: Keeps all errors, same as legacy behaviour.
  - `Reover.skipErrors`: Ignores all errors except seed errors. Used by new `ignoreErrors` flag.
  - `Recover.skipUpstreamErrors`: Ignores upstream errors but fails on combine errors. Currently unused.

## *Change 5*: Added unit tests to cover the new functionality.

Tests cover both `Signal` and `EventStream`, all variants of `reduce` and `scan`, and ~all combinations of the new `resetOnStop` and `ignoreErrors` flags~. Reorganised the existing `scan`-related tests. All new tests and existing `scan`-related tests pass, but there are `5` failures which are assumed to be unrelated.


src/main/scala/com/raquo/airstream/core/EventStream.scala
```diff
@@ -247,31 +250,18 @@ with DynamicImportStreamOps[A] // dynamicImport (Scala 3 only)
...
-    }
-  }
-
-  /** A signal that emits the accumulated value every time that the parent stream emits.
-    *
-    * @param fn Note: Must not throw!
-    */
-  def scanLeftRecover[B](initial: Try[B])(fn: (Try[B], Try[A]) => Try[B]): Signal[B] = {
-    new ScanLeftSignal(parent = this, () => initial, fn)
+  @inline final override def scanLeftRecover[B](
```
raquo: We only mark things final `final` if overriding them can break something.

For example, in `InternalTryObserver` the methods `onNext` and `onError` are final because the user is supposed to override `onTry`, if they override e.g. both `onNext` and `onTry` in a way that is inconsistent with each other, stuff will break.

I don't think `scanLeftRecover` is anything like that.

src/main/scala/com/raquo/airstream/scan/ScanLeftOps.scala
```diff
@@ -0,0 +1,117 @@
...
+   * @param combine     A binary operator that takes a tuple of the previously accumulated value and
+   *                    the next emission from this parent to produce the next accumulated value.
+   *                    It is safe for `combine` to throw uncaught exceptions, which are propagated through the error channel.
+   * @tparam B          The type of the accumulated value and thus of the resulting observable.
+   * @return            An [[Observable]] that emits the accumulated value every time this parent emits.
+   *                    The kind of observable produced is typically a signal or derivative thereof.
+   *
+   * @see [[scanLeftRecover]] has manual error handling, and [[reduceLeft]] has no `initial` seed value.
+   */
+  @inline final def scanLeft[B](
```
raquo: `@inline` should not be overused. Inlining blows up the bundle size when not used judiciously.

In general, things that are ok to `@inline`:
- Helper methods whose body is as small or smaller than their invocation (then it won't increase bundle size)
- Private helper methods that only have 1-2 uses, and were extracted for reading convenience rather than code reuse.

In general I would suggest not inlining stuff, there's too much other stuff to talk about.

src/main/scala/com/raquo/airstream/core/EventStream.scala
```diff
@@ -247,31 +250,18 @@ with DynamicImportStreamOps[A] // dynamicImport (Scala 3 only)
...
+    initial: Try[B],
+    resetOnStop: Boolean,
+  )(
+    combine: CombineTry[A, B],
```
raquo: As as a user if I see `CombineTry` in the signature, I have no idea what it is.

It's just a type alias for `(Try[A], Try[B]) => Try[B]`. This is typical type signature for a foldLeft-like method, it's not hard to understand in its original form. It is much more obvious than `CombineTry[A, B]`.

This indirection is not worth it IMO.

src/main/scala/com/raquo/airstream/scan/Recover.scala
```diff
@@ -0,0 +1,76 @@
...
+   *
+   * An event is silently ignored if it is a [[Failure]].
+   * The accumulator simply maintains its previous state until a successful event arrives.
+   */
+  @inline def skipUpstreamErrors[A, B]: RecoveryStrategy[A, B] = combine => {
+    case (Success(current), Success(next)) => Try(combine(current, next))
+    case (Failure(error), _) => Failure(error)
+    case (current, Failure(_)) => current
+  }
+}
```
raquo: This shouldn't exist at all. If user wants to ignore upstream errors, that's not part of `scan*` operators, that's `upstream.recoverIgnoreErrors.scanWhatever`

src/main/scala/com/raquo/airstream/scan/Recover.scala
```diff
@@ -0,0 +1,76 @@
...
+  }
+
+  /**
+   * A reduction recovery strategy that skips upstream and combine errors but not seed errors.
+   *
+   * An event is silently ignored if it is a [[Failure]] or if it causes `combine` to throw an exception.
+   * The accumulator simply maintains its previous state until a successful event arrives.
+   */
+  @inline def skipErrors[A, B]: RecoveryStrategy[A, B] = combine => {
+    case (Success(current), Success(next)) => Try(combine(current, next)).orElse(Success(current))
```
raquo: Skipping errors is fine, but swallowing them is not. If you have an error in your `combine`, there's virtually no situation in which you want to swallow it without warning. That would be debugging hell.

What I think should happen is, if `combine` throws an error (or if upstream emits an error), the scan observable should emit that error. When the scan observable receives a new upstream event, it should continue from its last non-error state. (EDIT: to clarify, all this is about the `skipErrors = true` case)

src/main/scala/com/raquo/airstream/scan/Recover.scala
```diff
@@ -0,0 +1,76 @@
...
+   * A strategy for handling errors in a reduction operation.
+   * There are three kinds of error that a strategy must consider:
+   *   - Upstream Errors: Those originating from the parent or another ancestor.
+   *   - Seed Errors: Those originating from computation of the initial seed value.
+   *   - Combine Errors: Those originating from the combine function.
+   *
+   * Mechanically, a strategy is a functional that transforms a simple [[Combine]] reduction
+   * operator with no error handling into a [[CombineTry]] variant that can handle errors.
+   */
+  type RecoveryStrategy[A, B] = Combine[A, B] => CombineTry[A, B]
```
raquo: These types are too complicated due to abstraction & indirection. `RecoveryStrategy` should be a simple trait similar to `FlattenStrategy`.

We don't plan on letting users provide custom recovery strategies that would make use of types A and B, so RecoveryStrategy should not have those types as params. It should be a simple trait like:

```scala
trait RecoveryStrategy {
  def something[A, B](f: (B, A) => B): (Try[B], Try[A]) => Try[B]
}
```

and then implemented two object-s: `KeepErrors` and `SkipErrors`.

Actually I would also rename it to `ScanRecoverStrategy` since it's specific to scan operators, as opposed to operators like `recoverWithTry`.

README.md
```diff
@@ -2608,23 +2608,24 @@ Airstream-provided Observer constructors that let you specify an error handling
 
 #### Other Error Handling Considerations
 
-* **`scanLeft`** operator is unable to proceed when encountering an error, so such an observable will enter a permanent error state if it encounters an error. You can not use the standard `recover` method to recover from this. You need to use `scanLeftRecover` instead of `scanLeft` to supply your error handling logic.
+* By default, reduction operators such as `scanLeft` are unable to proceed when encountering an error;
+  any unhandled exception will cause the resulting signal to enter a permanent error state.
+  You can not use the standard `recover` method to recover from this.
+  The solution is to pass the flag `skipErrors = true` (recommended),
+  or instead use `scanLeftRecover` for custom error handling.
+  @TODO Should `skipErrors` be `true` by default?
```
raquo, line 2616: Yes, but later. Maybe v19 will have it on already. But let's keep behaviour as-is for now. (EDIT: aside from `makeInitial` naming as mentioned elsewhere, that's ok to break if needed to cleanup the overloads)

src/main/scala/com/raquo/airstream/state/StrictSignal.scala
```diff
@@ -69,17 +71,19 @@ with DebugSignalOps[StrictSignal, A] {
     )
   }
 
-  override def scanLeftRecover[B](
-    makeInitial: Try[A] => Try[B]
+  @inline override def scanLeftRecover[B](
+    makeInitial: Try[A] => Try[B],
+    resetOnStop: Boolean,
```
raquo: I think let's go with `resetOnStop = false` default for all these operators, it would be in line with behvaiour of other streams & signals. Add the default only to public operators on observables, not the implementating classes. (EDIT: I made this comment before reading about the overload-with-defaults problem, see my suggested solution to that in another comment).

src/main/scala/com/raquo/airstream/scan/ScanLeftOps.scala
```diff
@@ -0,0 +1,117 @@
...
+   * @tparam B      The type of the accumulated value and thus of the resulting observable.
+   * @return        An [[Observable]] that emits the accumulated value every time this parent emits.
+   *                The kind of observable produced typically matches the kind of the parent observable.
+   *
+   * @see [[scanLeft]] has an explicit initial seed value.
+   */
+  // Note: This is provided so that users don't have to pass an empty parameter list.
+  @inline final def reduceLeft[B >: A](
+    combine: Combine[A, B],
+  ): Reduce[B] = {
```
raquo: The other overload to this is:

```scala
  def reduceLeft[B >: A](
    resetOnStop: Boolean = false,
    skipErrors: Boolean = false,
  )(
    combine: Combine[A, B],
  ): Reduce[B]
```

Hmmm I'm not sure if an optional argument list like this is a good pattern. I don't think I've seen that in any Scala library APIs, it feels weird. I think I would rather `resetOnStop` and `skipErrors` be added as second and third arguments after `combine`, in a single argument list. Seems like a lesser evil? Fewer overloads, and more obvious type signature. We expect those args to be used rarely.

src/main/scala/com/raquo/airstream/scan/ScanLeftSignalOps.scala
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  @inline final override def scanLeftRecover[B](
+    initial: Try[B],
+    resetOnStop: Boolean,
+  )(
+    combine: CombineTry[A, B],
+  ): Self[B] = {
+    scanLeftRecover[B](combine(initial, _), resetOnStop)(combine)
+  }
```
raquo: Hm, I didn't think that a version of `scanLeft` for Signals where the user can provide an `initial` value (vs `makeInitial`) would make sense, but this implementation where this `initial` value is immediately combined with the signal's current value seems reasonable. 👍

I wonder if the version with `initial: Try[B]` should have a slightly different operator name than the one with `makeInitial: Try[A] => Try[B]`. Seems like we're running into overloading trouble with default params because both of those use the same name.

I'm not sure though, because I can't think of a good name right now.

If we rename the signals-only scanLeft operators that have `makeInitial` to something different, like `scanLeftF` or `scanLeftWithInitialF`, can we solve the problem with overloads and default params? If so, let's do that (yes, that it will be a breaking naming change for users of these signal methods, I can accept that). I'm not sure about the best name for this version, but want to at least get the structure of methods and overloads right. I think it should be simplified quite a lot, including removals of optional-argument-list overloads.

src/main/scala/com/raquo/airstream/scan/ScanLeftOps.scala
```diff
@@ -0,0 +1,117 @@
...
+ * A base trait for [[Observable]]s with reduction operators such as `scanLeft` and `reduceLeft`.
+ *
+ * @tparam Scan   The kind of observable that is created by `scanLeft`.
+ * @tparam Reduce The kind of observable that is created by `reduceLeft`.
+ * @tparam A      The type of value emitted by this observable.
+ *
+ * @note Implementations must define `scanLeftRecover` and `reduceLeft`.
+ * @see  [[ScanLeftSignalOps]] and [[ScanLeftStreamOps]] for specialisations to signals and streams respectively.
+ */
+trait ScanLeftOps[+Scan[+B] <: Observable[B], +Reduce[+B] <: Observable[B], +A] {
```
raquo: Pls rename +Scan => +ScanSelf and +Reduce => +ReduceSelf for consistency with other similar *Ops traits, so that at least the `Self` part of the name is recognizable.

src/main/scala/com/raquo/airstream/scan/ScanLeftSignalOps.scala
```diff
@@ -0,0 +1,178 @@
...
+   *                    If `false`, errors will persist until restart.
+   * @param combine     A binary operator that takes a tuple of the previously accumulated value and
+   *                    the next emission from this parent to produce the next accumulated value.
+   *                    It is safe for `combine` to throw uncaught exceptions, which are propagated through the error channel.
+   * @tparam B          The type of the accumulated value and thus of the resulting signal.
+   * @return            A [[Signal]] that emits the accumulated value every time this parent emits.
+   *
+   * @see [[scanLeftRecover]] has manual error handling, and [[reduceLeft]] has no `initial` seed value.
+   */
+  @inline final def scanLeft[B](
```
raquo: missing `override` modifier. Please make sure to always use the modifier when you're overriding.

raquo (commented):
Thanks, that's a good start.

Please see requested changes in code comments.

Also keep in mind my comments about scaladoc in the other PR. TBH I didn't read any of the scaladoc in this PR, it's just too excessive.

I haven't properly read the tests yet due to their sheer volume, but that +2000 test diff is what's concerning me. Do we really need all those tests? At first glance some of them seem to be repetitive / redundant (again, I didn't actually read much of them, so it's a genuine question).

Like, `reduceLeftDefault` is a method with a two-line implementation. Does it really need what looks like 100 lines of tests? What are we testing in those tests, that isn't tested in the tests for `reduceLeftOption`? And what are we testing in `reduceLeftOption` tests that isn't tested in `scanLeft`. From a cursory glance, the two lines in `reduceLeftDefault` are pretty simple, they don't justify a 50:1 ratio of tests LOC to code LOC. Tests are not free. They take time to review, update (when desired behaviour changes), run, etc.

src/main/scala/com/raquo/airstream/scan/ScanLeftSignalOps.scala
SgtSwagrid: I think that's the right decision. I very much wanted to rename the existing `scanLeft` on signals before, but had assumed you wouldn't want a breaking change like that. The existing `scanLeft` on signals with `makeDefault` is kind of strange imo, and I would imagine that doesn't get too much use. Or maybe I just haven't met the use case yet. I'm glad you like the new version :)

Renaming the previous `scanLeft` entirely solves the problem with the overloads and the default args, which is great. For now, I went with the name `scanLeftGenerated`. Not terrible, not amazing, a little verbose, all I could come up with for now. Let me know if you think of anything better. My second-best idea was `scanLeftFrom`, but that's not particularly informative.

Would you like for me to also add `@deprecated` versions of the previous functions which redirect to the new ones? This could conflict a bit with the new functions with the same names is all.

src/main/scala/com/raquo/airstream/scan/Recover.scala
SgtSwagrid: Makes sense. This is the only comment that I haven't had a chance to address yet, but I'll get on that tomorrow.

src/main/scala/com/raquo/airstream/scan/ScanLeftSignalOps.scala
raquo: Somehow I've literally never used scanLeft with `makeInitial` myself. I agree that it's weird and it's hard for me to imagine that it's very popular, which is why I'm ok breaking the name. I don't think we need same-named deprecations for this.

Also, we will likely break not just the name but also the default skipErrors behaviour (in v19), which is a bigger issue as it's a behaviour change that users would need to test for, so I'm doubly not worried about the name change in comparison.

Let's keep the `scanLeftGenerated` name as a placeholder for now. It's easy to change later, even after merging. This PR will go into the v19 branch (yet to be created) so the temporary name is NBD.

src/main/scala/com/raquo/airstream/core/Signal.scala
```diff
@@ -132,27 +133,17 @@ with DynamicImportSignalOps[A] // dynamicImport (Scala 3 only)
...
-    *
-    * @param makeInitial currentParentValue => initialValue   Note: must not throw
-    * @param fn (currentValue, nextParentValue) => nextValue
-    * @return
-    */
-  override def scanLeftRecover[B](
-    makeInitial: Try[A] => Try[B]
+  override def scanLeftGeneratedRecover[B](
+    makeInitial: Try[A] => Try[B],
+    resetOnStop: Boolean,
```
raquo: Should default to `false`

src/main/scala/com/raquo/airstream/scan/ScanLeftStreamOps.scala
```diff
@@ -0,0 +1,78 @@
...
+  def reduceLeftDefault[B >: A](
+    default: => B,
+    resetOnStop: Boolean = false,
+    skipErrors: Boolean = false,
+  )(
+    combine: (B, A) => B,
+  ): Signal[B] = {
+    reduceLeftOption[B](combine, resetOnStop = resetOnStop, skipErrors = skipErrors)
+      .map(_.getOrElse(default))
+  }
```
raquo: I don't think this method should exist – it's just `scanLeft` under a different name (and with more limited type constraints), no?

src/main/scala/com/raquo/airstream/scan/ScanLeftStreamOps.scala
```diff
@@ -0,0 +1,78 @@
...
+  def reduceLeftOption[B >: A](
+    combine: (B, A) => B,
+    resetOnStop: Boolean = false,
+    skipErrors: Boolean = false,
+  ): Signal[Option[B]] = {
+    scanLeft[Option[B]](None, resetOnStop = resetOnStop, skipErrors = skipErrors) {
+      case (None, next) => Some(next)
+      case (Some(previous), next) => Some(combine(previous, next))
+    }
+  }
```
raquo: I don't think this method should exist. Seems to me that `stream.reduceLeft(...).startWithNone` would do the same thing? We shouldn't re-implement translation of `EventStream[A]` into `Signal[Option[A]]` in unrelated operators. Ideally operators should be orthogonal to each other where their functionality does not overlap, but they can be combined ergonomically.

raquo (commented):
Thanks, now that the overloads were cleaned up, I could see that some of the remaining methods are probably not needed. Please see attached code comments.

raquo:
I'm cautiously onboard with `stream.reduceLeft` returning a stateful stream, as you have it implemented now. Although in general I would rather avoid stateful streams, I think we can roll with it this once, and see how it pans out in practice.

In terms of naming, this presents a conundrum: traditional observable name for `reduceLeft` would be [scan](https://reactivex.io/documentation/operators/scan.html), however I don't feel great about `stream.scan` returning a stream while `stream.scanLeft` returns a signal. Well, maybe that's the lesser evil, I dunno. We can rename later.

src/main/scala/com/raquo/airstream/scan/ScanLeftStreamOps.scala
SgtSwagrid: This one I'm not particularly attached to either, and I'm happy to get rid of it.

That being said, the sematics are not identical to `scanLeft`, the difference being that `reduceLeftDefault` will not fold the seed value into the accumulator. After the first event, the seed disappears.

src/main/scala/com/raquo/airstream/scan/ScanLeftStreamOps.scala
SgtSwagrid: For this one, I would push back a little bit. In particular, `reduceLeft` is currently implemented in terms of this operator (not the other way around), and that is perhaps the most natural implementation of `reduceLeft`.

We had the question "should reduceLeft return a signal or a stream?", and there are downsides to both (signal is more natural for storing state, but then the value is wrapped in option, where for a stream this isn't necessary). Since there's no obvious best choice, it can't hurt to let the user decide upfront. An alternative naming scheme could be `reduceLeftToSignal` and `reduceLeftToStream`.

src/main/scala/com/raquo/airstream/scan/Recover.scala
SgtSwagrid: I'm thinking that instead of `(Try[B], Try[A]) => Try[B]`, the type for a general `merge` function in `scanLeftRecover` should be something more like `(Option[B], Try[A]) => (Option[B], Try[B])`, where the left output is for the accumulator and the right output gets emitted. Or even `(Option[B], Try[A]) => (Option[B], Option[Try[B]])`.

An alternative to this might be: `(Try[B], Option[B], Try[A]) => Try[B]`, where the new `Option[B]` is the last non-error value.

SgtSwagrid: Also with those semantics, a name like `continueOnError` might make more sense.

SgtSwagrid: However, in order to avoid changing any more stuff, I decided to add the `skipErrors` flag directly to `ScanLeftSignal`. Without changing the type of the combine-with-try function as above, it's impossible (as far as I can tell) to support your version through the combine-with-try function alone. This means that `RecoveryStrategy` is no longer needed at all, and also that `skipErrors` has to be added to the `XXXRecover` method variants as well (which isn't all bad, because to an extent the behaviour is orthogonal).

The new behaviour is:
When `skipErrors` is `true`, try to replace the LHS of the input to `combine` with the last known good value if there is one. If there has never been a good value, it will still be `Failure`, so the existing signature of `combine` is still meaningful.

src/main/scala/com/raquo/airstream/scan/ScanLeftStreamOps.scala
raquo: Hm, tbh that behaviour is kinda weird. I think let's nuke it. If anyone ends up asking for it (and `splitStart` does not fit their bill), we can review then, but we gotta be mindful of scope creep and maintenance burden.

src/main/scala/com/raquo/airstream/scan/ScanLeftStreamOps.scala
raquo: The downside is maintenance, and that's not just code maintenance, but maintaining a consistent API. Every time I add a new operator, I have to burn through a lot of calories making sure that its behaviour will be consistent with other operators. You just saw how much work that is in this very PR – and all the changes contained here are about scanLeft.

I prefer to err on the side of shipping fewer features with lower burden. I have to keep this sustainable.

`stream.reduceLeft(...).startWithNone` seems to do this job just fine so we can nuke this operator from the public API and nuke the tests that's just for it, that's a great deal for any maintainer.

I wouldn't let the implementation details of `reduceLeft` drive public API design. I have to give the concerns above more weight, because dealing with issues like that consumes a significant amount of my time.

Also keep in mind that it's sort of lucky that this Signal-based implementation even works. I thought we'd need to implement a ReduceStream class for that. Maybe we'll still need that in the future, I dunno.

src/main/scala/com/raquo/airstream/core/WritableSignal.scala
```diff
@@ -12,13 +12,29 @@ trait WritableSignal[A] extends Signal[A] with WritableObservable[A] {
 
   protected var maybeLastSeenCurrentValue: js.UndefOr[Try[A]] = js.undefined
 
+  protected var maybeLastSeenSuccessValue: js.UndefOr[A] = js.undefined
+
```
raquo: Let's definitely not mess with `core` logic, we don't need to. "lastSeenSuccessValue" is a scanLeft idea, we can implement it within `ScanLeftSignal`. Please roll back these `core` changes including `incrementUpdateId` / `removeCurrentValue` / `loadCurrentValueFromParent`.

I thought implementing `skipErrors` would be easier. To keep you sane I went and implemented it myself in the [scanleft-resumeOnError](https://github.com/raquo/Airstream/compare/master...scanleft-resumeOnError) branch. To keep myself sane, I did this based off `master`. Please `git cherry-pick` the one commit in that branch into this PR's branch. There will be some conflicts because you moved the ScanLeftSignal file, but if you rollback your recent `core` changes first, I think you can just take my version of ScanLeftSignal wholesale (aside from the `fn` -> `combine` rename, but that's easy). Be sure to carry over the other changes in my commit too, e.g. the tests and a couple minor edits in other files.

In that branch, I made `scanLeft` skip errors, whereas `scanLeftRecover` has current behaviour. Let's keep it like this, non-configurable. Fewer things to worry about.

Also I added a `tryOrFailure` helper to guard `Try => Try` functions in scalLeft against exceptions. Feel free to use that in your operators but please don't change other existing operators that are not related to this PR, this can be done separately later, I want to keep the scope from creeping.

raquo (commented):
Heyo, see new code comments below. Since it turned out that implementing skipErrors requires knowledge of core internals, I've implemented it in a separate commit myself to help out, please integrate that into your changes.

src/main/scala/com/raquo/airstream/core/WritableSignal.scala
raquo: > non-configurable

And let's remove the `resetOnStop` option too. Currently scanLeft behaves as `resetOnStop = false`, and that's good enough. It might have been a mistake to want this option for a signal (scanLeft). Signals always have a value, so the concept of resetting doesn't really make sense for them in general. It CAN make sense for scanLeft, because the operator itself contains an initial value that we could fall back to, but in general, resetting is a concept for streams in Airstream.

In the interest of reducing the amount of work left in this PR and especially to avoid going deeper into something that might not make sense, let's cut the `resetOnStop` option from the scope.

Sorry that I didn't see this initially.

I'm hoping that `resetOnStop = false` being the only option will work fine for the other operators like reduce. It seems that it should, given that it's the default throughout Airstream, but I guess we can always add it later if it's really needed.

src/main/scala/com/raquo/airstream/scan/ScanLeftOps.scala
```diff
@@ -0,0 +1,86 @@
...
+    resumeOnError: Boolean,
+  )(
+    combine: (Try[B], Try[A]) => Try[B],
+  ): ScanSelf[B]
+
+  /** Convert a reduction function into one that is error-aware and keeps all errors. */
+  protected def keepErrors[X, Y](combine: (Y, X) => Y): (Try[Y], Try[X]) => Try[Y] = {
+    case (Success(current), Success(next)) => Try(combine(current, next))
+    case (Failure(error), _) => Failure(error)
+    case (_, Failure(error)) => Failure(error)
```
raquo: `Try(combine(current.get, next.get))` is shorter and more efficient.

src/main/scala/com/raquo/airstream/core/WritableSignal.scala
```diff
@@ -5,9 +5,9 @@ import scala.scalajs.js
 import scala.util.{Failure, Success, Try}
 
 /** This trait exists to provide private functionality that can not be covariant in A.
-  * Concrete classes like MapSignal can extend this trait, but more general types like
-  * Signal[A] need to remain covariant in A for good ergonomics.
-  */
+ * Concrete classes like MapSignal can extend this trait, but more general types like
+ * Signal[A] need to remain covariant in A for good ergonomics.
+ */
```
raquo: ^ Please run scalafmtAll, it will clear this diff and more.

src/test/scala/com/raquo/airstream/state/LazyStrictSignalSpec.scala
```diff
@@ -483,7 +483,7 @@ class LazyStrictSignalSpec extends UnitSpec with BeforeAndAfter {
     val s =
       sourceVar.signal
         .map(identity)
-        .scanLeft(_.id) { (acc, foo) => acc + foo.id }
+        .scanLeftGenerated((foo: Foo) => foo.id) { (acc, foo) => acc + foo.id }
```
raquo: I see you've added type ascription here and elsewhere when using `scanLeftGenerated`. Did type inference break, to need the `: Foo` type ascription? It should be just `_.id`.

src/test/scala/com/raquo/airstream/misc/ScanLeftSignalSpec.scala
raquo, line 210: There's a merge conflict:
- I've added new tests to this file in my commit
- You've added tests to this file AND moved it into the `scan` package.

As a result, now we have two versions of this file. The tests that were in this file BEFORE me and you added our new tests are duplicated in both this file and your copy of it. For example the test "ScanLeftSignal made with Signal.scanLeft" is currently in both of these files.

To resolve the conflict, please remove (from your file!) the original tests that are duplicated in this file. To keep the diff readable, please do not change / move / rename this file, I'll move it myself after merging the PR eventually. Only change your new test files as needed to remove the duplication.

To be clear, I'm talking about removing actual duplicate tests, like "ScanLeftSignal made with EventStream.scanLeft", of which we have two copies now. I'm not sure if your **new** tests overlap with my **new** tests, if that's the case, I can live with it. Let's just get rid of the obvious duplicates.

raquo (commented):
Just minor stuff left – see comments.

Aside from that, there should be no outstanding issues, right? Seems fine to me.

Don't worry about the docs in readme, I'll update those later.

src/test/scala/com/raquo/airstream/state/LazyStrictSignalSpec.scala
SgtSwagrid: The type inference originally broke when using the variant with the overloaded function name, but now it's all better as we changed that.

SgtSwagrid:
Sorry for the delay!

Everything looks good to go from my side. Duplicate tests were removed, the diff looks clean, no conflicts, `keepErrors` is now as you suggested, unnecessary type ascription removed, tests pass, etc.

Thanks for helping me see this through. Next we should add `reduceRight`, which probably involves some kind of time travel.

SgtSwagrid:
@raquo I assume you're busy or on vacation or something. But in case this just managed to slip by (I had forgotten about it too), this is just a reminder that it exists. All good anyway, and no rush. Enjoy the sunshine ☀.

PR #157, opened 2026-09-15
Test: reproduce FetchStream lifecycle bugs

nguyenyou:
Adds two failing reproducers: stopping a fetch leaves its cancellation source subscribed, and restarting an abortOnStop fetch reuses an aborted signal.

Expected: unsubscribe on stop and use a fresh signal on restart. Tests only; no implementation changes.


raquo:
Nice, thanks. Fixed.

PR #158, opened 2026-09-15
Test: reproduce flattenMerge initial stream loss

nguyenyou:
Adds a failing reproducer for `Val(EventStream.fromValue(42)).flattenMerge`: expected `42`, but no event is emitted.

Tests only; no implementation changes.


raquo:
Thanks, good find!

PR #159, opened 2026-09-15
Test: cover sibling removal during owner cleanup

nguyenyou:
Adds a regression test where one subscription's cleanup kills a later subscription owned by the same owner.

Owner disposal should finish without throwing, clean each subscription exactly once, and leave the owner reusable.


raquo:
This test already passes, so I understand it to be additional coverage. Thanks! I've integrated it into the main test suite, and also generated a fuzzing test that tries to break the logic with a bunch of random subscription kills.

PR #160, opened 2026-09-15
Test: reproduce disabled transaction depth rejection

nguyenyou:
Adds a failing reproducer for `Transaction.maxDepth = -1`: expected nested transactions to execute, but both are rejected with `TransactionDepthExceeded`.

Tests only; no implementation changes.


raquo:
Thanks, good find. I fixed the boolean condition but I'll skip on adding a test for this because it basically requires us to blow the stack during the test run, and I don't feel good about doing that on purpose even if it's caught. Ultimately the maxDepth = -1 is just a debug helper that nobody noticed was broken for a while, and it seems unlikely that I'll screw up that same boolean again, so I'll leave it untested.

PR #161, opened 2026-09-18
Fix: splitMatchSeq handlers receive StrictSignal, as documented

nguyenyou:
`splitMatchSeq`'s docstring says handler signals are `StrictSignal`s "so you can read `.now()` from them", but `SplitMatchSeqContext` typed them as plain `Signal`. `Signal.now()` is `protected[airstream]`, so for users the documented call doesn't compile:

    method now cannot be accessed as a member of Signal[Int]

`SplitMatchSeqSpec` calls `.now()` in seq handlers and only compiles because it lives inside the `com.raquo.airstream` package.

The runtime child is already a `StrictSignal` (`splitSeq` yields a `KeyedStrictSignal`), and `splitMatchOne` handlers already receive one, so this makes the seq variant consistent with the rest of the family.

**Changes**
- `handleCase` / `handleType` / `handleRest` seq handlers: `Signal[_]` → `StrictSignal[_]`; same for the lambda the macro synthesizes for `handleValue`.
- Docstrings updated to match.

**Tests**
- `SplitMatchSeqStrictSignalRegressionSpec` — placed *outside* `com.raquo.airstream` so it sees what users see. Uses `assertCompiles`, so the regression shows as a failed test rather than a broken test build. Fails 3/4 before the fix, passes 5/5 after (incl. a runtime `.now()` test and a source-compatibility check for handlers written against `Signal`).
- `SplitMatchHandleValueOwnerRegressionSpec` — coverage for `handleValue` bodies containing local `val`/`def`/lambda definitions (the one place the macro moves user code into a lambda it synthesizes). Not a bug today (verified with `-Xcheck-macros`), but nothing exercised this shape.

Source-compatible: `Function1` is contravariant, so existing `(s: Signal[B]) => ...` handlers still typecheck.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


raquo:
Oh, that's my bad. I didn't review AI's macro refactor well enough since I don't really speak macros, and this slipped by. 

> SplitMatchSeqSpec calls .now() in seq handlers and only compiles because it lives inside the com.raquo.airstream package.

Fun. I wonder if it would be good practice to move the whole test suite into a non-airstream package, to catch these kinds of issues. I don't think I'm using any Airstream-private methods in the tests. Well, not routinely anyway.

Good find, thanks!

src/main/scala-3/com/raquo/airstream/split/MatchSeqClause.scala
```diff
@@ -16,7 +16,7 @@ sealed trait MatchSeqClause[+O]
   *
   * `I` is the element type of the collection being split. These methods are never evaluated – the
   * `splitMatchSeq` macro reads the clause shapes at compile time and generates a single match block.
-  * Handlers receive a plain `Signal[_]` (not a `StrictSignal[_]`), matching `splitMatchSeq`'s semantics.
+  * Handlers receive a [[StrictSignal]], like `splitSeq` and `splitMatchOne`, so `.now()` is available.
```
raquo, line 19: Lmao apparently I didn't even read the comments

PR #162, opened 2026-09-18
Perf: reduce split memoization allocations

nguyenyou:
Construct `SplitSignal`'s replacement memo entry directly instead of allocating an intermediate signal/output pair and mapped `Option`. Projection, key equality, and update behavior stay unchanged.

Local `fullOptJS` benchmarks with 2,000 keyed items showed about 12% less sampled allocation per collection update; retained heap was unchanged.

Validation: `scalafmtAll` run; full `sbt +test` passed with Node.js/jsdom: 891 tests on Scala 2.13 and 911 on Scala 3 (one ignored in each).


raquo:
Nice, thanks 👍

PR #163, opened 2026-09-19
Perf: avoid repeated combine readiness scans

nguyenyou:
Replace the full readiness scan on every parent event with a missing-input counter. Preserve failure handling, cached values across restarts, duplicate parent positions, and transaction batching; add regression coverage for these cases.

Local `fullOptJS` benchmarks in Chrome 153 (median of five rounds):

| Inputs | Start, emit first values, stop | Steady update |
| --- | --- | --- |
| 256 | 86.3 → 41.2 µs (52% less time) | 2.15 → 1.79 µs (16% less) |
| 2,000 | 2,875 → 332 µs (88% less) | 15.8 → 13.3 µs (16% less) |

Two-input performance was essentially unchanged. Synthetic measurements, not application-wide speedups.

Validation: `scalafmtAll`; full suites passed on Scala 2.13 and Scala 3 using Bun with jsdom browser APIs.


raquo:
Man you're on a roll. That's great, thanks. 👍

PS: our public API caps out at 22 inputs, I want to have a talk with anyone doing 2000 :)

PR #164, opened 2026-09-19
Perf: release cleanup callbacks after subscription teardown

nguyenyou:
Replace a subscription's cleanup callback with a shared no-op before invoking it, so retained dead handles no longer keep captured resources alive. Preserve the double-kill/reentrancy guard. The throwing-cleanup test is defensive coverage for misuse outside the documented contract.

Local `fullOptJS` heap benchmark in Chrome 153: retaining 2,000 killed observer subscriptions, each originally referencing a signal with a 1,024-element numeric array, retained **11.1 MB → 0.17 MB** after forced GC. Both individual and owner-driven cleanup showed the reduction. This measures retained dead handles, not total application heap. Lifecycle CPU measurements stayed within ±1%.

Validation: `scalafmtAll`; full `+test` on Scala 2.13 and Scala 3 using Bun with jsdom browser APIs.


raquo:
Nice 👍

Your benchmarks retain the references to the killed subscriptions, right?

I expect GC to collect everything if you don't explicitly retain references to subscriptions. If that's not the case, I would like to investigate further.

This is a good change either way, just want to make sure that there isn't a deeper/sibling issue where Subscription-s themselves are retained past our need for them.

nguyenyou:
Yes, the benchmark deliberately retains the killed subscriptions. Once those references are released, GC reclaims the payload even without this change. We saw no separate subscription-retention issue in these tests.

PR #165, opened 2026-09-19
Perf: reduce batched Var validation allocations

nguyenyou:
Use native `JsSet` to detect duplicate underlying Vars before batched writes, avoiding copied tuples and Scala-set allocations. Derived Vars still resolve to their underlying SourceVar. Uses the existing `ew` dependency; no persistent state or early-exit logic.

`JsSet` compares object identity. Built-in SourceVars already use identity equality; custom subclasses overriding `equals`/`hashCode` will now be compared by identity too.

Local `fullOptJS` benchmark: Scala 3.9.0, Scala.js 1.22.0, Chrome 153, macOS arm64. Repeated `Var.set` on unobserved Vars with preconstructed arguments; original upstream → this change:

| Vars/batch | µs/batch | Sampled bytes/batch |
| --- | --- | --- |
| 2 | 0.206 → 0.171 | 918 → 710 |
| 8 | 3.018 → 0.447 | 4,282 → 1,190 |
| 64 | 24.292 → 2.960 | 34,367 → 5,822 |
| 512 | 208.203 → 23.291 | 263,919 → 43,276 |

CPU: medians of six warmed rounds with rotated/reversed variant order. Allocation: medians of three rotated rounds after 10,000 warmup batches, measuring 20,000 batches (2,500 at size 512). Sampled allocation estimates, not retained heap; synthetic Chrome results, not application-wide claims.

Validation: `scalafmtAll`; full `+test` on Scala 2.13.18 and Scala 3.9.0 using Bun with jsdom browser APIs (399 + 428 tests passed).


raquo:
If you have the benchmark setup handy, could you please try collecting into a [JsSet](https://github.com/raquo/ew/blob/master/src/main/scala/com/raquo/ew/JsSet.scala)? Or into scalajs js.Set. Might need a bit of extra glue code, but I have a hunch that native JS structures could be more efficient. I try to use them for private structures.

nguyenyou:
Thanks for the tip! Tried `JsSet`—larger batches run ~88% faster with ~82% less allocation. Tests pass, and I’ve updated the PR with the results and a note about identity equality.

raquo:
Awesome. I've expanded this pattern to cover all previous usages of hasDuplicateTupleKeys as well.

PR #166, opened 2026-09-20
Perf: avoid Try allocation in FilterStream

nguyenyou:
`FilterStream` wrapped every predicate call in `Try(...).fold`, adding avoidable per-event overhead on the successful path.

Use a direct `try/catch NonFatal` around only the predicate. Downstream internal observer failures remain unhandled transaction errors, matching the previous behavior.

Benchmark on a clean optimized Scala.js `fullOptJS` bundle with Bun:

- 0.0517 → 0.0490 µs/event
- ~5% faster
- 3.2 KiB smaller optimized bundle (3,291 bytes)

Validation: focused regression test plus the existing Scala 2/3 test suites.


raquo:
It seems that an exception in `fireValue` would be previously emitted into `onError`, whereas now it would be thrown raw in `onNext`.

The test appears to validate that throwing in `onNext` causes the error to end up in unhandled-errors, but I'm actually not sure if that's strictly equivalent to previous behaviour in all cases. Seems like it could be happenstance, which I guess could be the reason for the test, but seems fragile.

I haven't analyzed this properly but OTOH it seems safer to keep behaviour strictly as-is, and since it's cumbersome to spell out, maybe it's worth extracting this into a reusable method in `util`, something like `tryFold(: => body)(onError, onSuccess)` and have errors from onSuccess flow into onError as well – basically it would work exactly same as Try(body).fold(...) except it wouldn't instantiate or return any Try-s.

Thoughts?

Also, I see that `WritableStream.fireValue` catches exceptions in externalObservers.foreach, but not internalObservers.foreach. InternalObserver.onNext says it must not throw, although that seems like a fragile guarantee too. We have a bunch of those in Airstream, including in functions that accept user code, which is not great, and needs a review eventually. But that is a lot more work to decide & verify all that behaviour, definitely out of scope for v18. So I'm inclined to keep existing behaviour when implementing small performance optimizations for v18.

nguyenyou:
I agree. The allocation win is small, while changing the exception boundary could have surprising consequences. I think it’s safer to leave this optimization out for now, so I’ll close the PR.

PR #168, opened 2026-09-20
Perf: specialize Vector findUpdate

nguyenyou:
Specialize `Splittable.findUpdate` for `Vector` using `indexWhere` and `updated`.

This stops predicate evaluation at the first matching item and returns the original Vector when there is no match, avoiding an unnecessary rebuild.

Benchmark: https://github.com/nguyenyou/airstream-findupdate-benchmark/actions/runs/35507833069

Screenshot 2026-09-20 at 18 28 42

Validation: 891 Scala 2 tests and 911 Scala 3 tests passed.


raquo:
Hm. On one hand it seems that this would be a nice optimization for all the indexable collections...

On the other hand:

> 24 KB larger benchmark bundle

OTOH that seems unrealistically high for one method. Is that difference consistent / reproducible on your end? Was that measured with fastLink by any chance?

nguyenyou:
Good catch — the +24 KB came from an unisolated benchmark bundle that included other changes. With only this Vector change, the delta is just **+1 byte** in both `fastOptJS` and `fullOptJS`. The performance improvement still reproduces.

raquo:
Huh. Well, I must say +1 byte is suspiciously **low**, but if you're sure that it's right, then all good.

nguyenyou:
Hmm, this is my bad — I trusted the AI response without verifying it carefully enough.

The 24 KB difference is real. I asked AI to build a small benchmark repo so we have a reproducible test that runs on GitHub Actions:

https://github.com/nguyenyou/airstream-findupdate-benchmark/actions/runs/35507833069

For the standalone `fullOptJS` bundles:

* Before: 121,511 bytes
* After: 145,721 bytes
* Difference: 24,210 bytes

I also broke down where the increase comes from:

* `indexWhere` itself: ~255 bytes
* Supporting iterator code (`sliceIterator`, `concat`, related iterators): ~3–4 KB
* Additional `Vector` update/iterator support: ~1–2 KB
* Generated bridges, lambdas, and type-related code: ~1 KB
* The remaining ~18–19 KB comes from shared Scala.js runtime and optimizer-generated code that gets pulled in or reshaped by those dependencies

So the 24 KB number is real for this isolated benchmark, but it should be described as **standalone linker overhead**. It does not mean that a real Laminar application will necessarily grow by 24 KB, since much of that runtime code may already be present in the application bundle.

raquo:
Hrm. I'll leave this open for now. Gotta make a release, and I'm not sure if the 24KB tradeoff is worth it. We can review later.

